PR #1, opened 2016-01-27
Add ability to pass extra variables on the command line.

aliafshar:


claat/export.go
```diff
@@ -115,11 +129,12 @@ func writeCodelab(dir string, clab *types.Codelab, ctx *types.Context) error {
...
-		Prefix:   ctx.Prefix,
-		GlobalGA: ctx.MainGA,
-		Meta:     &clab.Meta,
-		Steps:    clab.Steps,
+		Env:       ctx.Env,
+		Prefix:    ctx.Prefix,
+		GlobalGA:  ctx.MainGA,
+		Meta:      &clab.Meta,
+		Steps:     clab.Steps,
+		ExtraVars: parseAdditionalVars(),
```
crhym3: `writeCodelab` is called for each codelab when multiple are given on the input, so `parseAdditionalVars` would unmarshal the same data all over again, as many time as there are codelabs.

I would do it in the main, after `flag.Parse` and then just use it here.

claat/render/template.go
```diff
@@ -27,11 +27,12 @@ import (
...
-	Prefix   string
-	GlobalGA string
-	Meta     *types.Meta
-	Steps    []*types.Step
+	Env       string
+	Prefix    string
+	GlobalGA  string
+	Meta      *types.Meta
+	Steps     []*types.Step
+	ExtraVars map[string]string // Extra variables passed from the command line.
```
crhym3: Maybe just name it `Extra`? It's shorter :) and still makes sense to what it is, I think.

claat/main.go
```diff
@@ -33,6 +33,7 @@ var (
 	tmplout  = flag.String("f", "html", "output format")
 	prefix   = flag.String("prefix", "../../", "URL prefix for html format")
 	globalGA = flag.String("ga", "UA-49880327-14", "global Google Analytics account")
+	tmpldata = flag.String("data", "", "Additional arguments to pass to format templates. JSON object of string,string key values.")
```
crhym3: Also would rename this to `-extra` maybe?

crhym3:
thanks @aliafshar! could you also rebase this to the latest master? I've just fixed ci in https://github.com/googlecodelabs/tools/commit/6351d8e00f3c3f98d4a4bc402e3079c849d77871

claat/render/template.go
crhym3: Also, what about making it `map[string]interface{}`? That way you could pass strings, ints, floats, arrays and literally anything else that can be unmarshaled.

claat/export.go
aliafshar: Done.

claat/main.go
aliafshar: Done.

claat/render/template.go
aliafshar: Done making it Extra. I think map[string]string is fine. It is all we likely need ever, easier to use, and it is easy to update to interface if we ever do need it.

aliafshar:
https://github.com/googlecodelabs/tools/pull/2

Sorry, I couldn't understand github. :(

crhym3:
Closing in favour of #2.

PR #10, opened 2016-06-13
Add autospacing around italics in Markdown rendering.

chrisrecher:
This should stop the output from jamming subsequent Markdown parsing performed on the output. I suspect that the risk of accidental double spaces isn't a big problem, as once parsed as breaking spaces to HTML, they'll be consolidated into one.

If we should fight double spacing, I can trivially add a lookback mechanism similar to lineStart, but a lookahead mechanism would be a bit more architecturally complex. WDYT?


crhym3:
I think this LGTM for now. We can always go back to it and fix double spacing later on.
Thanks!

PR #23, opened 2017-02-09
Adds support for tables, infoboxes and youtube in MD files

petele:
I'd like to use this for WebFundamentals when we export code labs and other files

### DIFF
--- a/claat/parser/gdoc/parse.go
+++ b/claat/parser/gdoc/parse.go
@@ -729,10 +729,14 @@ func link(ds *docState) types.Node {
 	if ds.flags&fMakeCode != 0 || isCode(ds.css, ds.cur.Parent) {
 		t.Code = true
 	}
-	if href == "" || href[0] == '#' {
+	if href == "" {
 		t.MutateBlock(findBlockParent(ds.cur))
 		return t
 	}
+  if href[0] == '#' && href[1] == 'h' && href[2] == '.' {
+    t.MutateBlock(findBlockParent(ds.cur))
+    return t  
+  }
 
 	n := types.NewURLNode(href, t)
 	n.Name = nodeAttr(ds.cur, "name")
--- a/claat/render/md.go
+++ b/claat/render/md.go
@@ -17,6 +17,7 @@ package render
 import (
 	"bytes"
 	"fmt"
+  htmlTemplate "html/template"
 	"io"
 	"path"
 	"sort"
@@ -61,7 +62,11 @@ func (mw *mdWriter) writeString(s string) {
 }
 
 func (mw *mdWriter) writeFmt(f string, a ...interface{}) {
-	mw.writeString(fmt.Sprintf(f, a...))
+  mw.writeString(fmt.Sprintf(f, a...))
+}
+
+func (mw *mdWriter) writeEscape(s string) {
+  mw.writeString(htmlTemplate.HTMLEscapeString(s))
 }
 
 func (mw *mdWriter) space() {
@@ -115,7 +120,7 @@ func (mw *mdWriter) write(nodes ...types.Node) error {
 		case *types.InfoboxNode:
 			mw.infobox(n)
 		//case *types.SurveyNode:
-		//  mw.survey(n)
+		//	mw.survey(n)
 		case *types.HeaderNode:
 			mw.header(n)
 		case *types.YouTubeNode:
@@ -210,7 +215,7 @@ func (mw *mdWriter) list(n *types.ListNode) {
 }
 
 func (mw *mdWriter) itemsList(n *types.ItemsListNode) {
-	mw.newBlock()
+  mw.newBlock()
 	for i, item := range n.Items {
 		s := "* "
 		if n.Type() == types.NodeItemsList && n.Start > 0 {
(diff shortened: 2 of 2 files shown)


claat/render/md.go
```diff
@@ -232,3 +233,37 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
...
+		}
+		mw.writeString("\n</tr>")
+	}
+	mw.writeString("</table>")
+}
+
+func (mw *mdWriter) infobox(n *types.InfoboxNode) {
+	mw.newBlock()
+	mw.writeString(`<aside markdown="1" class="`)
+	mw.writeString(string(n.Kind))
```
x1ddos: there's also mw.writeBytes

claat/render/md.go
```diff
@@ -232,3 +233,37 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
...
+	mw.writeString("</aside>")
+}
+
+func (mw *mdWriter) youtube(n *types.YouTubeNode) {
+	mw.newBlock()
+	mw.writeString(`<div class="video-wrapper">\n`)
+	mw.writeString(`  <iframe class="devsite-embedded-youtube-video" `)
+	mw.writeString(`data-autohide="1" data-showinfo="0" frameborder="0" `)
+	mw.writeString(`allowfullscreen data-video-id="`)
+	mw.writeString(string(n.VideoID))
```
x1ddos: mw.writeBytes

claat/render/md.go
```diff
@@ -232,3 +233,37 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
...
+	mw.writeString(string(n.Kind))
+	mw.writeString(`">`)
+	mw.write(n.Content.Nodes...)
+	mw.writeString("</aside>")
+}
+
+func (mw *mdWriter) youtube(n *types.YouTubeNode) {
+	mw.newBlock()
+	mw.writeString(`<div class="video-wrapper">\n`)
+	mw.writeString(`  <iframe class="devsite-embedded-youtube-video" `)
```
x1ddos: not convinced by the "deviste-" prefix... would it be possible to drop? i.e. just `embedded-youtube-video`.

claat/render/md.go
x1ddos: better yet, merged with the line above:

```go
mw.writeFmt("data-video-id=%q", n.VideoID)
```

in which case I would also move "allowfullscreen" part to the line above it.

claat/render/md.go
x1ddos: Actually, you could just use `mw.writeFmt("<... class=%q>", n.Kind)`

But, because it's HTML now, you'll have to make sure it's safe using escaping. See this, for instance:
https://github.com/googlecodelabs/tools/blob/508f3fab2b82b08747277d730570444f1ab2c444/claat/render/html.go#L135

claat/render/md.go
```diff
@@ -159,11 +164,7 @@ func (mw *mdWriter) url(n *types.URLNode) {
 	if n.URL != "" {
 		mw.writeString("[")
 	}
-	for _, cn := range n.Content.Nodes {
-		if t, ok := cn.(*types.TextNode); ok {
-			mw.writeString(t.Value)
-		}
-	}
+	mw.write(n.Content.Nodes...)
```
x1ddos: Is this to make the "download" button a non-empty link text?

petele: Yep, otherwise it is just an empty block right now

claat/parser/gdoc/parse.go
```diff
@@ -729,10 +729,14 @@ func link(ds *docState) types.Node {
 	if ds.flags&fMakeCode != 0 || isCode(ds.css, ds.cur.Parent) {
 		t.Code = true
 	}
-	if href == "" || href[0] == '#' {
+	if href == "" {
 		t.MutateBlock(findBlockParent(ds.cur))
 		return t
 	}
+  if href[0] == '#' && href[1] == 'h' && href[2] == '.' {
```
x1ddos: why this change?

claat/render/md.go
```diff
@@ -209,7 +215,7 @@ func (mw *mdWriter) list(n *types.ListNode) {
 }
 
 func (mw *mdWriter) itemsList(n *types.ItemsListNode) {
-	mw.newBlock()
+  mw.newBlock()
```
x1ddos: this seems unnecessary.

claat/render/md.go
```diff
@@ -232,3 +238,33 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
...
+  mw.writeString(`<aside markdown="1" class="`)
+  mw.writeEscape(string(n.Kind))
+  mw.writeString(`">`)
+  mw.write(n.Content.Nodes...)
+  mw.writeString("</aside>")
+}
+
+func (mw *mdWriter) youtube(n *types.YouTubeNode) {
+  mw.newBlock()
+  mw.writeFmt("<iframe src=\"https://www.youtube.com/embed/%q\" ", n.VideoID)
```
x1ddos: 1. you can use backquotes. it reads better:

    mw.writeFmt(`<iframe src="..." `, n.VideoID)

2. the "%q" wraps string in quotes. I don't think it's what you want. you probably meant "%s"
3. are we sure n.VideoID is safe? if in doubt, I'd use [url.PathEscape](https://golang.org/pkg/net/url/#PathEscape)

claat/render/md.go
```diff
@@ -16,6 +16,8 @@ package render
 
 import (
 	"bytes"
+	"fmt"
+  htmlTemplate "html/template"
```
x1ddos: I think you can import it as is, without renaming: import ( "html/template" )
there's no other "template" in this file, is there?

claat/render/md.go
```diff
@@ -232,3 +238,33 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
...
+    mw.writeString("\n</tr>")
+  }
+  mw.writeString("</table>")
+}
+
+func (mw *mdWriter) infobox(n *types.InfoboxNode) {
+  mw.newBlock()
+  mw.writeString(`<aside markdown="1" class="`)
+  mw.writeEscape(string(n.Kind))
+  mw.writeString(`">`)
```
x1ddos: maybe the above three lines could be merged into one. seems simpler:

```go
mw.writeFmt(`<aside markdown="1" class="%s">`, template.HTMLEscapeString(n.Kind))
```

marcacohen:
Hi @petele . Any chance to follow through on this? If you're busy, I can run with it. I'd like to see this get added, especially the video embedding.

josephkandi:
What happened to this? Any chance this will go through?

claat/render/md.go
baoduy1103: relly ??

PR #25, opened 2017-02-20
Ensure we detect all codelabs step with md parser

didrocks:
The following structure:
 # Page title

 ## First Step
 Duration: 0:01

Will render this intermediate html output with blackfriday:
<h1>Page title</h1>

<h2>First Step</h2>

<p>Duration: 0:01</p>

The issue is the call to .advance() on the token, while treating h1, will then
be an empty string. Consequence is that inStepTitle is thus false and will
never have the opportunity to become true again (same with two steps separated
by a blank line).

The fix ensure that we can detect and reenter a new step, even if we went out
of any of them.


x1ddos:
This was @cassierecher so assigning it to him.

cassierecher:
her*

US holiday today; will take a look when I get back to the office tomorrow.

x1ddos:
sorry. thanks!

PR #26, opened 2017-02-20
Standardize the tags and append to the codelab field

didrocks:
Mirroring the gdoc where tags are coming from both Environments
and Tags, support the same independant tags for Markdown parser.


x1ddos:
Same here, asking @cassierecher to review.

cassierecher:
US holiday today; will take a look when I get back to the office tomorrow.

PR #28, opened 2017-03-28
[MD] Pass correct parameter to HtmlRenderer

didrocks:
blackfriday.HtmlRenderer is using HTML tags and not Markdown extensions as
parameters. Passing thus common HTML tags instead (matching commonHtmlFlags
from blackfriday).

The invalid matches were making images (index of HTML_SKIP_IMAGES was matching
EXTENSION_FENCED_CODE) skipped from Markdown.


x1ddos:
Maybe @cassierecher can review this.

cassierecher:
Good catch, thanks!

LGTM

PR #29, opened 2017-03-28
Enable local relative path copy of images.

didrocks:
As all images were imported from gdoc (and so remote), the case of local images
from markdown files weren't handled.
Ensure we can now fetch images remotely (from gdoc, drive or remote reference
from a markdown file) or locally. We compute relative path from the source
file itself for local images and reuse it when running "update".

### DIFF
--- a/claat/fetch.go
+++ b/claat/fetch.go
@@ -227,20 +227,42 @@ func fetchDriveFile(id string, nometa bool) (*resource, error) {
 
 var crcTable = crc64.MakeTable(crc64.ECMA)
 
-func slurpBytes(client *http.Client, dir, url string, n int) (string, error) {
-	res, err := retryGet(client, url, n)
+func slurpBytes(client *http.Client, codelabSrc, dir, imgURL string, n int) (string, error) {
+	// images can be local in Markdown cases or remote.
+	// Only proceed a simple copy on local reference.
+	var b []byte
+	var ext string
+	u, err := url.Parse(imgURL)
 	if err != nil {
 		return "", err
 	}
-	defer res.Body.Close()
-	b, err := ioutil.ReadAll(res.Body)
+	if u.Host == "" {
+		if imgURL, err = restrictPathToParent(imgURL, filepath.Dir(codelabSrc)); err != nil {
+			return "", err
+		}
+		b, err = ioutil.ReadFile(imgURL)
+		ext = filepath.Ext(imgURL)
+	} else {
+		b, err = slurpRemoteBytes(client, dir, imgURL, 5)
+		ext = ".png"
+	}
 	if err != nil {
 		return "", err
 	}
+
 	crc := crc64.Checksum(b, crcTable)
-	file := fmt.Sprintf("%x.png", crc)
+	file := fmt.Sprintf("%x%s", crc, ext)
 	dst := filepath.Join(dir, file)
 	return file, ioutil.WriteFile(dst, b, 0644)
+}
+
+func slurpRemoteBytes(client *http.Client, dir, url string, n int) ([]byte, error) {
+	res, err := retryGet(client, url, n)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	return ioutil.ReadAll(res.Body)
 }
 
 // retryGet tries to GET specified url up to n times.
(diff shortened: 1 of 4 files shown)


x1ddos:
@cassierecher what do you think about this one? not sure if it makes any difference to you, but maybe it does.

my personal concern is security. what if I create a markdown with an image referencing something like `/etc/passwd`.

didrocks:
`/etc/passwd` isn't an issue per itself, but I guess you meant something like `/etc/shadow` which has the hash passwords values. This one isn't readable to the world anyway. The `claat` tool will only have the permissions that the user allowed it to have. I understand your concern about exposing by error things from your system you don't want. (but for sensitive items, it means you gave the claat tool those priviledges).

If we want to prevent people from doing this, we can maybe we can filter by extension (or mimetype) to allow a certain list of allowed items? What do you think about this?

cassierecher:
@x1ddos The overall strategy and implementation LGTM, but I agree that there's potential security implications. I defer to your expertise on that; I haven't done much in-depth security work.

claat/fetch.go
```diff
@@ -227,22 +228,42 @@ func fetchDriveFile(id string, nometa bool) (*resource, error) {
...
+	var b []byte
+	var ext string
+	u, err := url.Parse(imgURL)
 	if err != nil {
 		return "", err
 	}
-	defer res.Body.Close()
-	b, err := ioutil.ReadAll(res.Body)
+	if u.Host == "" {
+		imgURL = path.Join(path.Dir(codelabSrc), imgURL)
```
x1ddos: how about prepending "/" to imgURL here, to make sure it doesn't go like "../../../etc/passwd" or any other sensitive content.

then, path.Join should take care of cleaning up the final result anyway.

didrocks: I think that restricting elements to a subpath from the codelab directory is a nice way to mitigate those issues.

However, prepending a / to imgURL will not achieve this (I just tried and checked golang stdlib): `path.Join` is cleaning each elements of the `Join` (apart from the first one), and thus, strips any prepended or appended /. Thus, `path.Join("bar", "/../foo")` is equivalent to `path.Join("bar", "../foo")`.
I'm going to use some absolute path method to ensure we capture the asset as being in the tree of the codelab directory.

didrocks:
In light of the previous element, I think that ensuring the asset is a children of the codelab directory should be enough, what do you think?
The issue with checking mimetype is that it's ok for most of the types, but svgs are thus "text/plain; charset=utf-8", which will require us either to not allow them (as all sensitive data will match the same mimetype) or make that check useless, thoughts?

x1ddos:
Sure, let's do just the path cleanup.
Also, note that you should probably be using `filepath`, not the `path` package, when working with file system.

didrocks:
Sounds ok! I did thus push the asset being constrained in the corresponding codelab {;sub-}directory.
Also, changed the path -> filepath (never noticed that filepath had a Join() method)

Tell me if you prefer that I squash that in a single commit.

claat/fetch.go
```diff
@@ -227,22 +228,49 @@ func fetchDriveFile(id string, nometa bool) (*resource, error) {
...
+	var b []byte
+	var ext string
+	u, err := url.Parse(imgURL)
 	if err != nil {
 		return "", err
 	}
-	defer res.Body.Close()
-	b, err := ioutil.ReadAll(res.Body)
+	if u.Host == "" {
+		cdir, err := filepath.Abs(path.Dir(codelabSrc))
```
x1ddos: I may sound a bit paranoid but it would be nice if you extracted this part, sanitizing local file path, into a separate function and wrote a test with all possible edge cases. I would also use `testing/quick` in addition to manual test cases.

claat/fetch.go
```diff
@@ -227,22 +228,49 @@ func fetchDriveFile(id string, nometa bool) (*resource, error) {
...
+		if err != nil {
+			return "", err
+		}
+		imgURL = filepath.Join(path.Dir(codelabSrc), imgURL)
```
x1ddos: This should be `filepath.Dir`, not `path.Dir`, right.
Same above.

didrocks:
The latest commits should address and include your suggestions of factoring this out.
It does add tests for the functions (both manual ones with edge cases and fuzzy testing).

Let me know if this is what you had in mind. I'm happy to squash the commits if desired.

claat/fetch_test.go
```diff
@@ -16,12 +16,17 @@ package main
...
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
+	"path/filepath"
+
+	"testing/quick"
```
x1ddos: Could you remove empty lines from the standard pkg imports? That way they will all line up nicely.

x1ddos:
@didrocks sorry to keep this PR waiting. Just one nit in the package imports in fetch_test.go and I think we can merge. 

Please, do squash commits if possible.
Thanks!

claat/fetch_test.go
didrocks: You're right, seems like my go-import doesn't always act as nicely as desired. Fixing it!

didrocks:
No worry at all! Maintaining myself a lot of opensource repo, I'm know what this means on your side :)

Made the latest fix, resquashed in one commit (updating its description) and rerun the tests. All should be good now! Thanks for the review :)

x1ddos:
Thanks!

PR #35, opened 2018-01-30
claat: enable support for Team Drive

### LINKED ISSUE #34: Make claat command work with Team Drive
Team Drive has a slightly different semantics and requires clients to opt-in:
https://developers.google.com/drive/v3/web/enable-teamdrives

x1ddos:
We don't do anything specific, just fetching a couple metadata fields.
So, this seems to be as simple as opting-in as described in
https://developers.google.com/drive/v3/web/enable-teamdrives

Closes #34.


marcacohen:
LGTM

PR #37, opened 2018-02-18
complete readme with full instructions, background, and other info

marcacohen:


README.md
```diff
@@ -1,13 +1,162 @@
-# tools
+# Tools for authoring and serving codelabs
+Marc Cohen, [mcohen.io](https://mcohen.io), [@marcacohen](https://twitter.com/user/marcacohen)
```
marcacohen: @x1ddos I feel like you belong here more than me. Would it be ok if I include your info here as well?

marcacohen: btw, I think that gif is too heavy. I'll try making a smaller one.

README.md
```diff
@@ -1,13 +1,163 @@
...
+Codelabs are interactive instructional tutorials, which can be authored in Google Docs
+using some simple formatting conventions. You can also author codelabs using markdown syntax.
+This repo contains all the tools and documentation you’ll need for building and publishing
+your own codelabs.
 
 If you're interested in authoring codelabs, join [this group](https://groups.google.com/forum/#!forum/codelab-authors),
-which connects you with other authors and provides access to the 
-[Codelab Formatting Guide](https://docs.google.com/document/d/18dnMdUJQaGKY1Tit_-fO1YOpOpAbA4hh0YDXQlCEjvA/edit?ts=574ec5d9).
+which connects you with other authors and provides access to the [Codelab Formatting Guide](https://docs.google.com/document/d/18dnMdUJQaGKY1Tit_-fO1YOpOpAbA4hh0YDXQlCEjvA/edit?ts=574ec5d9).
+
```
x1ddos: Replace the doc URL with https://g.co/codelabs/guide shortcut? I saw it down below.

README.md
```diff
@@ -1,13 +1,163 @@
...
+1. Make one or more changes and preview your codelab, using the preview app provided by Google. To preview a codelab, install the Preview Codelab Chrome extension in your browser. Now you can preview a codelab directly from the Google Doc view by clicking the Chrome extension’s button, which will open a new tab to display the preview. Alternatively, navigate manually to https://codelabs-preview.appspot.com/?file_id=<google-doc-id>
+
+1. Install the claat command -- see the [README in the claat directory](https://github.com/googlecodelabs/tools/blob/master/claat/README.md) of this repo for instructions..
+
+1. Run the claat command to transform the doc contents into one of the supported output formats. The default supported formats are html and markdown but the claat tool supports adding additional formats by specifying a path to a Go template. For example, using the example document above:
+
+    `$ claat export -f html 1rpHleSSeY-MJZ8JvncvYA8CFqlnlcrW8-a4uEaqizPY`
```
x1ddos: "-f html" is the default. maybe drop it?

README.md
```diff
@@ -1,13 +1,163 @@
...
+    <br>`ok      your-first-pwapp`
```
x1ddos: hm... why is there a `<br>`?
if you use 4 spaces indentation, without back quotes, you shouldn't need it

README.md
```diff
@@ -1,13 +1,163 @@
...
+
+1. Run the claat serve command.
+
+    `$ claat serve`
```
x1ddos: remove back quotes. 4 spaces indentation already does it.

README.md
```diff
@@ -1,13 +1,163 @@
...
+The output generated by claat is a purely static set of html or markdown code. As such,
+it can be served by any web serving mechanism, including any of the following options:
+
+* Github Pages (*.github.io)
+* App Engine
+* Firebase Static Serving
+* Google Cloud Storage
+* Amazon S3
+* Any open source web server (Nginx, Apache)
+* Python -m SimpleHttpServer
```
x1ddos: I think you meant to use `-` instead of `*`.

README.md
marcacohen: fixed.

README.md
marcacohen: fixed

README.md
marcacohen: That doesn't seem to be the case with the vscode markdown previewer (which is why I added the <br>) but I'll try it and see if the GH renderer handles this better.

README.md
marcacohen: Same disp as previous.

README.md
marcacohen: * and - are interchangeable, at least per Gruber's original md proposal. Perhaps CommonMark has standardized this?

marcacohen: CommonMark says: A bullet list marker is a -, +, or * character.
I'm happy to switch if you prefer hyphens.

PR #38, opened 2018-02-18
claat serve: grey background and MS Windows fixes

x1ddos:
This is a follow up of #36.

- make sure all file paths manipulations are compatible with Windows
- clarify bower override map: repo resolution vs version override
- grey background in HTML template


claat/serve.go
```diff
@@ -72,214 +60,232 @@ func cmdServe() {
...
-		warned = true
-	}
-	out, err := os.Create(filepath)
-	if err != nil {
-		return err
-	}
-	defer out.Close()
-	resp, err := http.Get(url)
-	if err != nil {
+func writeFile(name string, content []byte) error {
```
marcacohen: Would it be more idiomatic to pass a Reader interface, rather than a byte string?

claat/serve.go
```diff
@@ -72,214 +60,232 @@ func cmdServe() {
...
+	"polymerelements/paper-scroll-header-panel":     "1.0.16",
+	"polymerelements/paper-styles":                  "1.2.0",
+	"polymerelements/paper-tabs":                    "1.8.0",
+	"polymerelements/paper-toast":                   "1.3.0",
+	"polymerelements/paper-toolbar":                 "1.1.7",
+	"polymerelements/platinum-sw":                   "1.2.4",
+	"polymerelements/sinonjs":                       "1.17.1",
+	"polymerelements/stacky":                        "1.3.2",
+}
+
```
marcacohen: Add short comment to say what this function does?

claat/serve.go
```diff
@@ -72,214 +60,232 @@ func cmdServe() {
...
+	for _, u := range tryURL {
+		if err := downloadFile(zipFile, u); err == nil {
+			ok = true
+			break
 		}
 	}
-	err = unzip(zipFile, basedir)
-	if err != nil {
+	if !ok {
+		return fmt.Errorf("fetchRepo(%q, %q): could not find in any of %q", name, spec, tryURL)
```
marcacohen: I know %q escapes strings but will it do the right thing in terms of formatting tryURL, which is array of strings? Or would you want to use %v for that arg?

claat/serve.go
x1ddos: Done

claat/serve.go
```diff
@@ -72,214 +60,232 @@ func cmdServe() {
...
 		Dependencies map[string]string
 	}
-	err = json.Unmarshal(raw, &b)
+	if err := json.Unmarshal(raw, &bower); err != nil {
+		return fmt.Errorf("%s: %v", bowerFile, err)
+	}
+
+	for name, spec := range bower.Dependencies {
+		if err := fetchRepo(basedir, name, spec); err != nil {
+		}
```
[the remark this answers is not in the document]
x1ddos: Done

x1ddos:
Marc said LGTM :)

PR #44, opened 2018-04-09
add support for image title and alt attributes

marcacohen:


claat/render/html.go
```diff
@@ -160,6 +160,12 @@ func (hw *htmlWriter) text(n *types.TextNode) {
 
 func (hw *htmlWriter) image(n *types.ImageNode) {
 	hw.writeString("<img")
+	if n.Alt != "" {
+		hw.writeFmt(` alt="%s"`, n.Alt)
```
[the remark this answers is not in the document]
marcacohen: fixed

marcacohen:
@x1ddos Thx for quick review, comments fixed, PTAL.

claat/parser/md/parse.go
```diff
@@ -406,11 +406,23 @@ func handleInfobox(ps *parserState) {
...
 func handleImage(ps *parserState) {
+	var n *types.ImageNode
+	var alt, title string
 	for _, v := range ps.t.Attr {
 		if v.Key == "src" {
-			ps.emit(types.NewImageNode(v.Val))
-			break
+			n = types.NewImageNode(v.Val)
+		}
+		if v.Key == "alt" {
```
x1ddos: I'd rather use switch, which won't go into all the checks like these `if`s do. Also, note `strings.ToLower`:

```go
switch strings.ToLower(v.Key) {
case "src":
  n = types.NewImageNode(v.Val)
case "alt":
  alt = v.Val
case "title":
  title = v.Val
}
```

claat/render/md.go
```diff
@@ -148,9 +148,18 @@ func (mw *mdWriter) text(n *types.TextNode) {
...
+		mw.writeString(n.Alt)
+	} else {
+		mw.writeString(path.Base(n.Src))
+	}
 	mw.writeString("](")
 	mw.writeString(n.Src)
+	if n.Title != "" {
+		mw.writeString("\"")
+		mw.writeString(n.Title)
+		mw.writeString("\"")
```
x1ddos: I think you're missing a space before `"title text"`. Also, probably safer to use quoted string:

```go
if n.Title != "" {
  mw.writeString(fmt.Sprintf(" %q", n.Title))
}
```

marcacohen: fixed

PR #51, opened 2018-04-24
claat: update command line help message

x1ddos:
Because we keep modifying built-in templates,
I thought it would be good to warn users.


marcacohen:
LGTM, thanks for adding

PR #33, opened 2018-01-18
Implement new md parser

didrocks:
Here is the md parser we are using for quite almost a year in ubuntu. (powering https://tutorial.ubuntu.com).

It's inspired by the google doc parser. I didn't propose it beforehand as I wasn't given time to rewrite the md test suite and I think we should have an interface and share some code between the 2 parsers.
However, @samtstern convinced me to put this PR up against the upstream repo and will handle the necessary tests additions. ;)

You can see a scaffold of a markdown tutorial with syntax description here: https://github.com/canonical-websites/tutorials.ubuntu.com/blob/master/examples/example-tutorial.md (and see that it formats well with any markdown parser, like github).

Note as well that I wrote https://github.com/Ubuntu/tutorial-deployment (with tests) to generate an offline tutorial website (compatible with our web team site). The feature you can find interesting there is the possibility to save any markdown file and have the current tutorial web page refreshed (a little bit like the online gdoc refresh google appengine service for codelabs) to easily view your modifications. It only needs to use this polymer component: https://github.com/canonical-websites/tutorials.ubuntu.com/blob/master/src/elements/websocket-reloader.html which websocket signal and reload the page if this is the currently modified tutorial.

Hope that helps!

PR #53, opened 2018-06-19
ci: switch Go version to latest release; lint

x1ddos:
Previously, we used "master" version of Go on Travis CI.
Apparently, this uses a tip version and not the latest release.

So, use a fixed version.
The downside is we'll need to update it with each new Go release,
but it's still better than building release binaries using
a tip version.

Once submitted, https://github.com/googlecodelabs/tools/pull/33#issuecomment-398311679 can be rebased to make that PR green.

R: @didrocks @samtstern

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 sudo: false
+git:
+  depth: 3
 language: go
 go:
-- 1.8.x # need for App Engine
-- master
+- 1.8.x  # need for App Engine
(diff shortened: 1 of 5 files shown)


didrocks:
Looks correct, I still build on master for some of the project. Sad that Travis doesn't have a `latest` which is latest release version or that you can't put a config in "warning only" mode.

I'll rebase once merged on master.

PR #65, opened 2018-06-22
Markdown export improvements

### LINKED ISSUE #63: export/md: missing metadata
If you do the following:

```shell
$ claat export -f md $GDOC_ID
```

You get an MD file with no metadata section, the first line is a `#` tag.  So when you go to `export` the `md` file to get servable HTML you get:

```shell
$ claat export index.md 
err	index.md invalid metadata format, missing at least id: map[]
```

### LINKED ISSUE #64: export/md: durations are turned into sentences
I exported a Google Doc codelab to markdown:

```shell
$ claat export -f md 1YZj0tKwwRz1I8uJpW2J1FApD_a02Tmw8gWZvq-thfes
```

The markdown is mostly well-formed, but durations become italicized sentences:

```md
## Set up the sample project

*Duration is 5 min*
```

samtstern:
Fixes #63 
Fixes #64


claat/render/template.md
```diff
@@ -1,10 +1,21 @@
+---
+id: {{.Meta.ID}}
+{{if .Meta.Status}}status: {{.Meta.Status}}{{end}}
+{{if .Meta.Summary}}summary: {{.Meta.Summary}}{{end}}
+{{if .Meta.Author}}author: {{.Meta.Author}}{{end}}
+{{if .Meta.Categories}}categories: {{commaSep .Meta.Categories}}{{end}}
+{{if .Meta.Tags}}tags: {{commaSep .Meta.Tags}}{{end}}
+{{if .Meta.Feedback}}feedback link: {{.Meta.Feedback}}{{end}}
+{{if .Meta.GA}}analytics account: {{.Meta.GA}}{{end}}
```
x1ddos: I wonder if we should just always output these regardless of whether they're empty.
Then it's 1) easier to ready and 2) don't need to remember a field name when modifying existing markdown codelab.

But I'm ok with either. Whatever you prefer.

claat/render/md.go
```diff
@@ -108,8 +108,8 @@ func (mw *mdWriter) write(nodes ...types.Node) error {
 			mw.itemsList(n)
 		//case *types.GridNode:
 		//	mw.grid(n)
-		//case *types.InfoboxNode:
-		//	mw.infobox(n)
+		case *types.InfoboxNode:
+			mw.infobox(n)
 		//case *types.SurveyNode:
 		//	mw.survey(n)
```
x1ddos: I guess we should finish those commented out at some point too :)

claat/render/template.md
samtstern: Will leave it for now since we don't have tests to see how the parser deals with empty input on the metadata.

PR #69, opened 2018-07-02
Markdown bash rendering fix

josephch405:
Technically, terminal snippets should use bash code block rendering. Could cause issues when tools rely on the language annotation for terminal commands.


x1ddos:
@josephch405 we have 2 kinds of code blocks: code snippets and command line snippets.
Although I agree, command line snippets are techincally bash commands most of the time, they are not the same as bash code snippets.

This change would break all the codelabs which use command line snippets, not necessarily all in bash, turning them into code snippets with language "bash".

PR #76, opened 2018-07-27
Refactored ImportNodes and ImageNodes to be filtered in types library

josephch405:


josephch405:
Changes no longer needed after discussion

Issue #78, opened 2018-08-12
Create a claat build command


Gillespie59:
Hi, 

First I want to thank you for your tool. I am started using it for my own codelabs. 

If I understand well, the basic workflow is : 

```
claat export ...
claat serve 
```

The last command is important because all bower dependencies will be downloaded. 

Do we have another command for downloading these dependencies, without launching an HTTP server. For example a `claat build` ?

Second question, I am looking for a docker image in order to launch claat commands without Go on my system. Do you have this image ?

Thanks again.

x1ddos:
> Second question, I am looking for a docker image in order to launch claat commands without Go on my system. Do you have this image ?

You don't necessary need Go, neither Docker. See first paragraph in https://github.com/googlecodelabs/tools/blob/master/claat/README.md#install

PR #81, opened 2018-08-16
feat: create a new build command

Gillespie59:
Fixes https://github.com/googlecodelabs/tools/issues/78.


samtstern:
@x1ddos I believe this was in reference to https://github.com/googlecodelabs/tools/issues/78

Not sure about the `install` name but I think it's a good idea to have a command that gets the codelab ready for serving but without dictating/starting the server.  I could see this being use for CI or static builds.

x1ddos:
Ah ok. That makes more sense. How about actually using the proposed `build` command? Sounds closer to that it's supposed to do.

Either way, one thing missing from this change is a help section in main.

claat/serve.go
```diff
@@ -15,41 +15,15 @@
...
-	}
-	// Go get the dependencies.
-	if err := fetchRepo(depsDir, "google-codelab-elements", "googlecodelabs/codelab-components#2.0.2"); err != nil {
-		fatalf(err.Error())
-	}
-	if err := writeFile(filepath.Join("elements", "codelab.html"), codelabElem); err != nil {
-		fatalf(err.Error())
-	}
-
+	cmdInstall()
```
x1ddos: This is now `cmdBuild()`.

claat/main.go
```diff
@@ -180,6 +181,10 @@ When writing to a directory, existing files will be overwritten.
 
 The program exits with non-zero code if at least one src could not be exported.
 
+## Build command
+
+Install all the dependencies needed by the codelab.
```
x1ddos: Let's add something like:

> Install all the ... unless already installed.
> This is done automatically by the serve command.
> To clean up and rebuild all the dependencies, remove bower_components
> directory and run the build command again.

claat/main.go
```diff
@@ -188,6 +193,8 @@ Clicking on a directory representing an exported codelab will load
 all the required dependencies and render the generated codelab as
 it would appear in production.
 
+The serve command run first the build command.
```
x1ddos: "The serve command always runs the build command first." or something like that. I'm not a native. @samtstern please correct.

samtstern: @x1ddos yeah your suggestion LGTM

PR #88, opened 2018-08-24
Update Makefile to work even if the user doesn't have GOPATH set.

### LINKED ISSUE #87: `make` doesn't work if the user has not set GOPATH
GOPATH is now optional and users shouldn't be required to set it.

cassierecher:
This fixes #87


cassierecher:
How's this?

marcacohen:
LGTM, thanks Cassie

PR #89, opened 2018-08-25
Move commands to non-main subpackage.

cassierecher:
This makes it easier to call CLaaT commands as subroutines in other
programs.

### DIFF
--- a/claat/main.go
+++ b/claat/main.go
@@ -36,9 +36,6 @@ import (
 
 var (
 	version string // set by linker -X
-)
-
-var (
 	// commands contains all valid subcommands, e.g. "claat export".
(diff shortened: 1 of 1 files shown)


x1ddos:
Funny, I see no changes in this pull request :)

image

cassierecher:
Ah, I think I made a mistake and pushed to master instead of my fork. I like the idea of restricting `master`, how do we do that?

x1ddos:
@cassierecher my only comment on the submitted changes is that the flags to remain in the `main` package. If there are functions which use flag values directly, convert them to take arguments instead and pass the flag values that way.

Thanks!

cassierecher:
Addressed in #108

PR #104, opened 2018-10-18
Clean up redundant functions

cassierecher:


cassierecher:
The Travis error seems to be due to issues with the lint package/

PR #117, opened 2018-11-21
PoC: Use codelab-elements instead of codelab-components.

shawnbuso:
This needs some more work, but for now it's a PoC.

What this does:
- Stop using codelab-components
- Use codelab-elements instead by pulling a [release](https://github.com/shawnbuso/codelab-elements/releases/tag/0.1) from my fork of codelab-elements
  - This release was created by cloning the repo, running `bazel build :bundle`, then manually creating a release and uploading the bundle zip.

What I did:
- Update the build command to download codelab-elements instead of codelab-components
- Update the HTML template to use the new custom elements in place of the old polymer components.

What still needs to be done:
- Add the option to build and use a local copy of codelab-elements instead of pulling the released version
- Update the non-HTML templates to use custom elements
- Automate the release process for codelab-elements described above

The second point above should probably be done in this PR to avoid breakage, but I wanted to get this out for comments before I put a ton more time into it.

There's a *lot* of deleted code here - that's because the codelab-elements bundle includes its dependencies, so we don't need to do all the work of downloading bower dependencies like we did before.

The changes to the template are based on a diff of a codelab on the preview server (using codelab-elements) and the same codelab as output by the current CLaaT tool (using codelab-components).


arkassay:
I think we need a transition plan for the tools using the different libraries? We still have to finish fixing some bugs before me move permanently to custom elements, wdyt?

shawnbuso:
I wasn't totally sure on the relationship between codelab-components and this repo. Grep tells me the only place codelab-components and its dependencies are used in this repo is to serve the test server (`claat serve`), but I could be wrong.

PR #134, opened 2019-01-02
Adding bazel build to the Travis presubmit.

shawnbuso:


shawnbuso:
Removing reviewers for now until I verify that this is an OK practice.

.travis.yml
```diff
@@ -1,15 +1,25 @@
 sudo: false
+dist: trusty
 git:
   depth: 3
 language: go
 go:
 - 1.9.x  # need for App Engine
 - 1.11.x # latest release because "master" is dev version
+env:
+- BAZEL=0.19.1
```
nicolasgarnier: Should be 0.18.1 right?

nicolasgarnier:
In general I would say it's Good Practice to have the build running as a presubmit :)

shawnbuso:
Hah yea - Cassie mentioned potential concerns with the way we're installing bazel on Travis with each presubmit run, but [this is how the bazel folks themselves do it](https://github.com/bazelbuild/rules_go/blob/master/.travis.yml) so I would say it's the recommended method.

.travis.yml
shawnbuso: Good catch - fixed.

nicolasgarnier:
When you resolve the conflicts I think we could merge :)

PR #138, opened 2019-01-10
Change export command to receive sources from opts, not args list.

cassierecher:
I need this urgently, so I'm just changing the one command for now. I'll circle back and make the other commands not directly access flags.Args later.

### DIFF
--- a/claat/main.go
+++ b/claat/main.go
@@ -77,6 +77,7 @@ func main() {
 			GlobalGA:  *globalGA,
 			Output:    *output,
 			Prefix:    *prefix,
+			Srcs:      flag.Args(),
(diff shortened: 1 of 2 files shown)


claat/main.go
```diff
@@ -77,6 +77,7 @@ func main() {
 			GlobalGA:  *globalGA,
 			Output:    *output,
 			Prefix:    *prefix,
+			Srcs:      flag.Args(),
```
marcacohen: Do you need to store the srcs here for some other dependency? I do like the var change from args to srcs but why not just use:

srcs = unique(flag.args())

and then use srcs from that point forward but continue using a temp variable.

PR #140, opened 2019-01-12
Using a URL parser to get the hash from drawer elements when clicked.…

### LINKED ISSUE #139: Update handleDrawerClick_ to use a real URL parser to get the hash value
Currently handleDrawerClick_ calculates the step that was clicked by just removing the first character of the href, and assuming everything else is the step. This breaks if those links turn into full URLs to the step. For example, in the left nav:
```
<a href="#1"></a>
```
correctly navigates you to step 1, but:
```
<a href="codelabs.com/my-codelab#1"></a>
```
tries to navigate to step "odelabs.com/my-codelab#1". We should instead use a URL parser to parse the href param, then get the hash value from said URL parser.

shawnbuso:
… Fixes #139.

### DIFF
--- a/codelab-elements/google-codelab/google_codelab.js
+++ b/codelab-elements/google-codelab/google_codelab.js
@@ -486,7 +486,7 @@ class Codelab extends HTMLElement {
     }
 
     const targetUrl = new URL(target.getAttribute('href'));
-    // Substring bexcause the hash property includes the "#".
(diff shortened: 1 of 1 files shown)


shawnbuso:
Bump :)

shawnbuso:
Closing without merge since this was usurped by #158

PR #141, opened 2019-01-12
Parse language for code snippets in markdown.

anuraaga:
While the [README](https://github.com/googlecodelabs/tools/tree/master/claat/parser/md) mentions adding a code language as a hint for syntax highlighting, currently the language is not actually parsed out of the markdown. This change makes sure the language is propagated into the `CodeNode`.


claat/parser/md/parse.go
```diff
@@ -590,6 +590,16 @@ func code(ds *docState, term bool) types.Node {
 		v = "\n" + v
 	}
 	n := types.NewCodeNode(v, term)
+	for _, attr := range ds.cur.Attr {
+		if attr.Key == "class" {
+			for _, class := range strings.Split(attr.Val, " ") {
+				if strings.HasPrefix(class, "language-") {
+					n.Lang = class
```
narphorium: Wouldn't you want it without the "language-" prefix?

```suggestion
					n.Lang = class[len("language-"):]
```

claat/parser/md/parse_test.go
```diff
@@ -157,3 +157,22 @@ feedback link: https://www.google.com
 		t.Errorf("\ngot:\n%+v\nwant:\n%+v", c.Meta, wantMeta)
 	}
 }
+
+func TestParseCode(t *testing.T) {
+	wantCode := types.NewCodeNode("class Foo {}", false)
+	wantCode.Lang = "language-java"
```
narphorium:
```suggestion
	wantCode.Lang = "java"
```

claat/parser/md/parse.go
anuraaga: Hi @narphorium thanks for the review and sorry for the very late response. I didn't strip out the prefix here since I noticed here

https://github.com/googlecodelabs/tools/blob/46f2ab2d8b563f6f3134d9dc0ba6245a8c68dfec/claat/render/html.go#L221

`Lang` is added as a class without adding a prefix. The `google-prettify` library expects e.g., `language-java` as the class

https://github.com/google/code-prettify#how-do-i-specify-the-language-of-my-code

I wasn't sure it was appropriate to change the `render` method, mainly because I don't know what the semantics of `language=` should be I'm not aware of any language that respects that attribute.

Let me know if it's better to strip the prefix here, and update render to add it back for the class.

anuraaga: /cc @DanSanche

dmattia:
Friendly Ping

marcacohen (approved):
thanks!

anuraaga:
Looks like language support was added in #463. Nice that it finally landed.

PR #151, opened 2019-02-04
add support for storing the built artifacts in a globally readable place

marcacohen:


codelab-elements/google-codelab-analytics/google_codelab_analytics.js
```diff
@@ -294,7 +294,7 @@ class CodelabAnalytics extends HTMLElement {
     /** @type {!HTMLScriptElement} */
     const resource = /** @type {!HTMLScriptElement} */ (
         document.createElement('script'));
-    resource.src = '//www.google-analytics.com/analytics.js';
+    resource.src = 'https://www.google-analytics.com/analytics.js';
```
shawnbuso: Does this not auto-redirect to https when left as just "//"? This could cause issues for folks using browsers that don't support HTTPS (which I think unfortunately is still a thing).

marcacohen: good catch -- I made this change because this was yielding 404s but I realized this is because I was serving the site via file open. This works fine w/ // when I use a simple web server so reverting this URL change. Thanks!

marcacohen: PTAL

marcacohen: @shawnbuso PTAL? Also same for the related PR, thanks.

PR #155, opened 2019-02-07
added language to CodeNode constructor

DanSanche:
The README for the[ parser page](https://github.com/googlecodelabs/tools/tree/master/claat/parser/md) says the following:

> Put the name of the code language after the first fence to explicitly specify which highlighting plan to use.

This feature wasn't actually implemented, as the parser wasn't setting the language field in the constructed CodeNode. This PR fixes that


DanSanche:
Closing, just realized there was an [older PR](https://github.com/googlecodelabs/tools/pull/141) that does the same thing (and added a test)

PR #157, opened 2019-02-12
switch parser to blackfriday.v2

teeler:
Heyo!

I'm upgrading some other downstream dependencies on this repo to blackfriday.v2, so upgrading this repo while I'm at it, hope thats OK!

Ping me if you want to discuss (rtp@google.com).


teeler:
We should probably chat before you merge this...there are some logistical issues

PR #160, opened 2019-02-16
Replace `{{` by `&#123;&#123;`

### LINKED ISSUE #159: Need to escape {{ because of devsite
When uploading a codelab HTML to devsite, if the codelab contains double curly bracket `{{` there will be an error when trying to publish ro stage the codelab because 

Solution: we need to replace curly brackets with their character entity: `&#123;`

nicolasgarnier:
fixes #159

### DIFF
--- a/claat/render/html.go
+++ b/claat/render/html.go
@@ -12,328 +12,305 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package render
+package cmd
 
 import (
-	"bytes"
+	"encoding/json"
 	"fmt"
-	htmlTemplate "html/template"
-	"io"
-	"sort"
-	"strconv"
-	"strings"
+	"io/ioutil"
+	"log"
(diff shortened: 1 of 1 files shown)


nicolasgarnier:
Please squash commits when merging :)

claat/render/html.go
```diff
@@ -132,7 +137,8 @@ func (hw *htmlWriter) writeFmt(f string, a ...interface{}) {
 }
 
 func (hw *htmlWriter) writeEscape(s string) {
-	htmlTemplate.HTMLEscape(hw.w, []byte(s))
+	s = htmlTemplate.HTMLEscapeString(s)
```
skarEE: Do you want = or := ?

nicolasgarnier: `=` since `s` is already defined (I was getting a compile error with `:=`)?

PR #165, opened 2019-02-18
Fixing closure type declarations

nicolasgarnier:


nicolasgarnier:
Please squash commits before merge.

PR #166, opened 2019-02-18
CSS fix for icons inside buttons

nicolasgarnier:


nicolasgarnier:
That's in order to fix this: 
image

PR #180, opened 2019-03-06
Adding instructions for Youtube Videos.

nicolasgarnier:
But let's only merge once we have the previewer re-deployed and the publisher deployed with the custom elements. @shawnbuso to give us the green light on this.

### DIFF
--- a/FORMAT-GUIDE.md
+++ b/FORMAT-GUIDE.md
@@ -15,7 +15,7 @@ To preview a codelab:
 
 ## Formatting Reference
 
-1. Table of Contents    
+1. Table of Contents
 
     Every codelab document must use the **Heading 1** paragraph style to delineate the steps of the codelab. In turn, these headings will be used to automatically create a Table of Contents which shows the student exactly where they are in the codelab and lets them jump to any step.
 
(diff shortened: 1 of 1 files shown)


FORMAT-GUIDE.md
```diff
@@ -55,6 +55,13 @@ You can also use this to target specific events, for instance:  \
 
     Inline images in your codelab should just work seamlessly. You can re-size them in your codelab document and that width will be applied as a **max-width** on the image in the codelab markup so that images are the same size relative to the text but also scale down appropriately for smaller browsers. 
 
+1. Youtube Videos
+
+    Youtube Videos can be embedded by doing:
+     - Add an image in the document. The image can be a screenshot of the video for instance but it really doesn't really matter since it won't be displayed but replaced by the embedded video.
```
marcacohen: Two reallys are better than one?  :)

FORMAT-GUIDE.md
```diff
@@ -55,6 +55,13 @@ You can also use this to target specific events, for instance:  \
...
+     - Add an "Alt Text" to the image by doing **Cmd+Opt+Y** or **Right click > "Alt Text..."**
+     - Put a Youtube video link in the **Description** field of the Alt Text. in the format `https://www.yoututbe.com/watch?v=[video_ID]`
```
marcacohen: Can you include a start time?  Would be good to state whether that's supported or not.

nicolasgarnier: Currently it's not. But let's file a feature request. I can add that in another PR.

FORMAT-GUIDE.md
nicolasgarnier: Fixed! 🤣

FORMAT-GUIDE.md
marcacohen: Can you state here that it's not currently supported (so it's clear)?

marcacohen: I realize the set of things not supported is infinite :) but this is an obvious one I think people will wonder about.

nicolasgarnier: Done!

marcacohen: Thanks!

PR #187, opened 2019-03-07
Slight build changes to allow SCSS imports into devsite

nicolasgarnier:
- Split the global styles into their own files
 - Build target fixes


nicolasgarnier:
All tested and no issues!

PR #190, opened 2019-03-10
Center images and Youtube videos.

### LINKED ISSUE #114: We should center images
We need to find a way to center images. In many cases, it would make the generated codelab look much better. In fact, I would argue that it should be the default.

nicolasgarnier:
fixes #114 

This makes sure only images that are on their own lines are centered. And won't do anything to images that are inline text or if there are multiple images on the same line.

caveat: it adds a bunch of `<span></span>` to html output.

Example result:
image

image

### DIFF
--- a/claat/render/html.go
+++ b/claat/render/html.go
@@ -151,9 +151,6 @@ func (hw *htmlWriter) text(n *types.TextNode) {
 	if n.Code {
 		hw.writeString("<code>")
 	}
-	if !n.Bold && !n.Italic && !n.Code {
-		hw.writeString("<span>")
-	}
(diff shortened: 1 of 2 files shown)


nicolasgarnier:
Hold on I'll be changing this implementation since the issue here is that the existing codelabs that don't get regenerated may have issues with the new CSS.

nicolasgarnier:
Done. PTAL. instead of adding <span> everywhere this adds a class `image-container` to containers that only contain images and we can center the content in these containers.

nicolasgarnier:
Merging. FYI authors will have to republish their codelabs if they want the images to be centered (once we'll have updated the publisher)

PR #191, opened 2019-03-10
Using variable rather than hard coded Analytics ID

nicolasgarnier:


nicolasgarnier:
OK I just added the Codelabs level one then which was missing from the html template. With that template only the Global one was being tracked.

I don't see anywhere in the code where we could provide a category one though so it could be a regression of switching to the custom elements.

@shawnbuso could you do another release of claat and another redeploy of the publisher? This way codelabs that gets republished will get their codelabs specific tracking.

PR #192, opened 2019-03-20
Adding an "About codelab" custom element to the devsite template.

nicolasgarnier:
Adding an "About codelab" custom element to the devsite template.

This included the tasks:
 - Refactoring: renaming the "author" attributes as "authors" throughout. Low risk since Author was not really used before (only on tests) and was not surfaced in templates.
 - Passing the "Last Updated" date to the template context
 - Injecting the about element inside the first step and made sure it stays before, and outside, the instructions.

Also added the devsite-badger element.


claat/render/template-devsite.html
```diff
@@ -48,6 +49,9 @@
                     title="{{.Meta.Title}}"
                     environment="{{index .Env}}"
                     feedback-link="{{.Meta.Feedback}}">
+      <google-codelab-about {{if .Updated}}last-updated="{{.Updated}}"{{end}}
```
shawnbuso: Does it make sense for this to be in a template in the OSS repo if the element itself isn't in this repo?

nicolasgarnier: Not sure but my thinking was:
 - I'd like to avoid injecting elements dynamically when possible (that's what we'd be doing if that's only on devsite) since that makes the site (a bit) slower. Maybe we don't care that much given the amount of dynamic stuff going on though :)
- Imo i'd like to keep as much stuff in the open-source repo as possible. (We could even add that element to the version on codelabs.develoeprs.google.com. I think it would make some sense.)
- An empty element simply doesn't display anything if it has not been defined so it's kinda "backward compatible" in a sense. It doesn't really hurt having the element here.

nicolasgarnier: Also there is a JS change to do in the codelabs step element and it would not be super elegant if I had to do it from another location.

PR #229, opened 2019-04-25
Template changes

nicolasgarnier:
Added original_source meta tag to the devsite template and the codelab-about section to the html template.


claat/cmd/export.go
```diff
@@ -106,8 +106,8 @@ func exportCodelab(src string, opts CmdExportOptions) (*types.Meta, error) {
 	// codelab export context
 	lastmod := types.ContextTime(clab.mod)
 	meta := &clab.Meta
```
marcacohen: maybe more general (in case it ever needs to be used elsewhere) to do this:

clab.Meta.src = src

and then anything that uses clab.Meta will have access to the orig src doc.

nicolasgarnier: Done :)

PR #243, opened 2019-05-14
Material icon 'cloud_download' replaces Polymer icon 'file-download'

### LINKED ISSUE #242: Intent-to-implement: Show a 'file download' icon in custom-elements buttons
The 'file-download' icon () is no longer displaying on download buttons 
() due to it being a Polymer icon. We no longer use Polymer icons, and to maintain parity a material-icon should be chosen instead.

Similar icons:
- https://material.io/tools/icons/?icon=arrow_downward&style=sharp
- https://material.io/tools/icons/?icon=cloud_download&style=sharp

Marcial1234:
Fixes #242 
Download buttons now show as: 

Attempted to refactor 'button' function to use 'text/template' instead of 'if's to apply to style, but couldn't get 'stringifyNode' to work across packages. Will leave that for later.


claat/render/html.go
```diff
@@ -214,7 +214,7 @@ func (hw *htmlWriter) button(n *types.ButtonNode) {
 	}
 	hw.writeBytes(greaterThan)
 	if n.Download {
-		hw.writeString(`<iron-icon icon="file-download"></iron-icon>`)
+		hw.writeString(`<iron-icon icon="cloud_download"></iron-icon>`)
```
Marcial1234: ``

Marcial1234: https://material.io/tools/icons/?icon=cloud_download

nicolasgarnier:
@Marcial1234, we can also do this via a CSS fix which would be a bit better because if would apply to all existing codelabs whereas with a compiler fix we would need authors to republish their codelabs. Here is the fix we used in devsite:

```
iron-icon[icon="file-download"]::after {
  content: 'cloud_download';
}
```

Could you add that to https://github.com/googlecodelabs/tools/blob/master/codelab-elements/google-codelab-step/google_codelab_step.scss instead?

nicolasgarnier:
Actually we could have both fixes too :)

Marcial1234:
This PR doesn't seem to have fixed the issue. The cloud icon is not rendering since it's not in its correct form: `<i class="material-icons">cloud_download</i>`

@nicolasgarnier, your fix seems to only prepend 'cloud_download' as a plain string to Download buttons. We could revert this PR if needed.

Will need to send a PR fixing/revering this + correctly implementing #242 + copy the CSS styling for `iron-icon` now for `i`, which will be part of #264 as well yet still maintain backways compatibility.

Issue #247, opened 2019-05-20
Refactor repo to allow for modular use. Replace Node struct with Protobuf


Marcial1234:
Tracking issue.

Marcial1234:
Conversation that need to happen:

API changes:
- Links
  - GitHub 'code location' links are headings+links with prefixed favicons. They should instead work as the other prefixed favicon links (StackOverflow, Android, Google Cloud, Support)
  - Inconsistent CSS features
    - Underline on hover, not applicable to GitHub links
    - Blue font for Cloud and Support links, black font for Github links
    - Github links are bold
- Proposed changes:
  - *General*: Prefix all links of this kinds with favicons all the time, and not only under a "FAQ" section. Generalize look-and-feel features for all types, allow user to customize as they link (bold, italics, or both)
  - *GDoc*: Do not use *Headings 3* for GitHub links. Leave as plain links



Other elements
- Surveys
  - Does anyone still use them? If so, how? Were they replaced by analytics + feedback links?

- Fragments
  - Does anyone use them? HTML queries cannot be written to find these

Marcial1234:
Closed all pre design review PRs

Marcial1234:
Closing as change in direction of product.

PR #250, opened 2019-05-20
[proto-definitions - Part 1] Deleting Unused HorizontalRule

Marcial1234:
Part of #247

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -47,7 +47,8 @@ message BlockContent {
     CodeBlock code_block = 4;
     Table table = 6;
 
-    reserved 5, "horizontal_rule";
+    reserved 5; 
+    reserved "horizontal_rule";
   }
 };
 
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -45,13 +45,10 @@ message BlockContent {
     Image image = 2;
     InfoBox info_box = 3;
     CodeBlock code_block = 4;
-    HorizontalRule horizontal_rule = 5;
     Table table = 6;
-  }
-};
 
-message HorizontalRule {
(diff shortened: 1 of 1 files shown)


third_party/tutorial.proto
```diff
@@ -45,15 +45,10 @@ message BlockContent {
     Image image = 2;
     InfoBox info_box = 3;
     CodeBlock code_block = 4;
-    HorizontalRule horizontal_rule = 5;
```
cassierecher: [Reserve](https://developers.google.com/protocol-buffers/docs/proto#reserved) these.

Marcial1234: Done

third_party/tutorial.proto
```diff
@@ -46,6 +46,8 @@ message BlockContent {
     InfoBox info_box = 3;
     CodeBlock code_block = 4;
     Table table = 6;
+
+    reserved 5;
```
cassierecher: Also reserve the name "horizontal_rule"

Marcial1234: Done

third_party/tutorial.proto
```diff
@@ -45,13 +45,10 @@ message BlockContent {
...
     InfoBox info_box = 3;
     CodeBlock code_block = 4;
-    HorizontalRule horizontal_rule = 5;
     Table table = 6;
-  }
-};
 
-message HorizontalRule {
-  // No fields.
+    reserved 5, "horizontal_rule";
```
cassierecher: I think it's not possible to combine numbers and names. Try:

reserved 5;
reserved "horizontal_rule";

Marcial1234: You are correct.

PR #251, opened 2019-05-20
[proto-definitions - Part 2] Deleting Table from proto

Marcial1234:
Part of #247
Deleting `Table` Type as it's unused.

`CodeBlock`s, `InfoBox`es, and `Survey`s are the only elements that use "Table" as formatting in GDocs, but this format does not carry over to Markdown. Deleting to have a cleaner 1=>1 mapping of Types. 

Ideally, `Step` contents will be `repeated BlockContent`, and `BlockContent` will have all possible types that require special processing.

In a Future PR (#257) `Surveys` will be introduced.

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -45,6 +45,8 @@ message BlockContent {
     Image image = 2;
     InfoBox info_box = 3;
     CodeBlock code_block = 4;
+
+    reserved 5, 6;
   }
 };
 
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -1,0 +1,123 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


Marcial1234:
> Tables are a UI element, as with a Google Doc table. 
Correct, but in terms of rendering, we are no longer operating in GDoc space. The logic to detect tables and such stay on the parsing side, but we do not need to store these as is own abstract type as we're not rendering MD tables.

As an example, this is how code blocks are currently rendered:
https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/types/node.go#L287
https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/render/html.go#L226
https://github.com/googlecodelabs/tools/blob/879b5d21e27d54350c3b983d24e9584f4afc2971/claat/render/md.go

InfoBoxes follow a similar logic.

Marcial1234:
turns out we were supporting tables, but they were barely found in any codelab and there was no sample image on the authoring guide. Possible contents were updated as such.

PR #253, opened 2019-05-21
[proto-definitions - Part 3] No more special type for titles

Marcial1234:
Part of #247
Titles for are strings with different formatting for each parent-type. Does not deserve own subtype since it's a custom mapping for each (`Tutorial`, `Step`).

I'm attempting to maintain one proto-type per individual component.

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -17,19 +17,19 @@ syntax = "proto3";
 package devrel.tutorial;
 
 message Tutorial {
-  Title title = 1;
+  string title = 4;
   map<string, string> metadata = 2;
   repeated Step steps = 3;
+
+  reserved 1;
 };
 
 message Step {
-  Title title = 1;
+  string title = 4;
   map<string, string> metadata = 2;
   repeated Content content = 3;
-};
 
-message Title {
-  repeated InlineContent inline_contents = 1;
+  reserved 1;
 };
 
 message Content {


cassierecher:
By doing this we would lose special formatting in titles. I'm not sure if that's something we should support in the first place. I can understand something like italics. Adding Marc for further discussion.

cassierecher:
@marcacohen

marcacohen:
So this would lose special formatting of titles when codelabs are manipulated with protobufs, but would leave the existing functionality intact for the claat command? (i.e. if I used bold or italics in a title and am just using the claat command, not protobufs, will it still work?)

Marcial1234:
@marcacohen: The protobuf-based flow will ideally replace the current renderer completely, and all Codelab features will be able to be expanded from them.

These two are used independently so they shouldn't share a type (1), and they are already used as plain strings, per template files and struct definitions:

Codelab title: https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/render/template-devsite.html#L53
https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/types/meta.go#L76

Step title: https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/render/template-devsite.html#L53
https://github.com/googlecodelabs/tools/blob/f843e022fdda43cd8c9ebef69a5303d8e4771a0a/claat/types/meta.go#L76

I do not see where we can customize each to be bold &| italic.

Marcial1234:
The only place I'd advocating to remove bold and italics altogether are from Headings, per #254. But it's easily changeable to allow for them. Discussion for that can be moved there.

marcacohen:
ok, sounds like a non-issue then

PR #254, opened 2019-05-21
[proto-definitions - Part 4] Replace `Heading` content type to a plain string

Marcial1234:
Part of #247
Having special formatting (italics, bold) on Headings seems like overkill. Making their content as a plain string. The case of GitHub links that are formatted as Headings will be changed on the parser side, but these do not follow true/normal Headings rules.

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -69,7 +69,7 @@ message Heading {
   string text = 1;
   int32 level = 2;
 
-  reserved 1;
+  reserved 1, "inline_contents";
 };
 
 message Image {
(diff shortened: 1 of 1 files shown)


Marcial1234:
cc @marcacohen

third_party/tutorial.proto
```diff
@@ -66,8 +66,11 @@ message InlineContent {
 };
 
 message Heading {
-  repeated InlineContent inline_contents = 1;
+  string text = 3;
```
shawnbuso: Should be listed in numerical order by index. i.e., flip these lines.

Marcial1234:
Closing as too nit

PR #256, opened 2019-05-21
[proto-definitions - Part 6] Remove code_location from CodeBlocks

Marcial1234:
Part of #247
`code_location`: `CodeBlock`s and its code location are not tightly coupled.  In a future PR, this will be represented as a styled `Link` since it follows a similar styling to the `FAQ` styled icons (`code_location` links are links to bold links to GitHub with a prepended favicon, same as we have for StackOverflow, GCloud, etc)

`inline_contents`: I am not aware of how the automating syntax highlighting is done, but unless specified (font text is `Courier New` on GDoc), there is no special parsing done to the text/content of the table. Making this a string for such reason.

PR #258, opened 2019-05-21
[proto-definitions - Part 8] BlockContent is now main type of Steps

Marcial1234:
Part of #247
Refactor: Types are now shows as they appear by Depth-First (immediately define new types that show up, go back once a chain is done, repeat until done).

In a Future PR, repeated `InlineContent`s will be merged into a new type, called `Paragraph` (ideas on this welcomed), which will be added to `BlockContent`, and will replace other `repeated InlineContent`.

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -27,16 +27,9 @@ message Tutorial {
 message Step {
   string title = 4;
   map<string, string> metadata = 2;
-  repeated Content content = 3;
+  repeated BlockContent content = 5;
 
-  reserved 1;
-};
-
-message Content {
-  oneof content {
-    InlineContent inline_content = 1;
-    BlockContent block_content = 2;
-  }
+  reserved 1, 3;
 };
 
 message BlockContent {
@@ -55,11 +48,6 @@ message BlockContent {
   }
 };
 
-message Link {
-  string text = 1;
-  string location = 2;
-};
-
 message InlineContent {
   oneof content {
     string text = 1;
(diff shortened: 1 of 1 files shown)


Marcial1234:
Per offline discussion with Cassie, decided against creating a dedicated field that agglomerates `InlineContent`.

third_party/tutorial.proto
```diff
@@ -27,16 +27,9 @@ message Tutorial {
...
 
-  reserved 1;
-};
-
-message Content {
-  oneof content {
-    InlineContent inline_content = 1;
-    BlockContent block_content = 2;
-  }
+  reserved 1, 3;
```
shawnbuso: 3?

shawnbuso: Add reserved names

third_party/tutorial.proto
```diff
@@ -46,14 +39,9 @@ message BlockContent {
...
     List list = 7;
-
-    reserved 5, 6;
   }
-};
 
-message Link {
-  string text = 1;
-  string location = 2;
+  reserved 5, 6;
```
shawnbuso: Why are 5 and 6 reserved here?

third_party/tutorial.proto
```diff
@@ -111,3 +98,9 @@ message List {
   }
   ListVariety list_variety = 2;
 };
+
+message ListStep {
+  repeated InlineContent contents = 2;
+
+  reserved 1;
```
shawnbuso: What's 1?

third_party/tutorial.proto
Marcial1234: 3 was `content` with type `Content`, which is deleted on this PR. reusing name.

third_party/tutorial.proto
Marcial1234: addressed: https://github.com/googlecodelabs/tools/pull/256/files#r289119654

PR #260, opened 2019-05-22
add gulp based landing page generator

marcacohen:
this gives user the ability to use the "site" tool - a gulp directory which allows you to generate a default landing page along with custom views

don't be too afraid of the size here -- it's replicated internal code, and a ton of vendored dependencies. The main job was that I had to de-googleize everything. Let me know if you see anything I missed.


nicolasgarnier:
Usually it's common practice not to check-in all the dependencies but instead, check-in a **bower.json** file that will reference all the dependencies and add a setup step to run `bower install` to pull all the dependencies. Would that work in you case?

marcacohen:
I thought about that but we just removed the Polymer dependency and I'm reluctant to add a bower dep. I don't love forcing people to install bower, which is becoming increasingly less popular it seems. I prefer just giving them everything they need to get up and running, at the cost of a much bigger download. I'm open to talking it over if you disagree.

site/README.md
```diff
@@ -0,0 +1,354 @@
...
+      // Optional: If true, do not include this view in the list of views on
+      // the home page. It will still be accessible via the direct URL.
+      "hidden": false,
+    }
+    ```
+
+1. (Optional) Add a file named `style.css` inside the view folder. If provided,
+this file will be included in the HTML, allowing for custom styles.
+
+    This file is not included in the main assets bundle, so there will be a
```
sethvargo: This might be less relevant in the public-facing README

nicolasgarnier:
I agree bower is not that popular anymore. It looks like most of these deps are available on npm though :D perhaps we could transition to npm. I'm OK if we do that later on in another PR though.

sethvargo:
FWIW, I tried to transition to npm when I did this internally, but it was a sadness. There's transitive dependencies in the npm manifest that conflict with some of the components and it turned into a giant rabbit hole.

marcacohen:
I'd love to get rid of bower altogether but given Seth's experience, I
think I'll defer that step to a future PR.

Marcial1234:
What was the context behind this feature?

marcacohen:
Yes thanks for the cross-ref

PR #263, opened 2019-05-23
[proto-build-system Part 1] Added protobuf building to current build system

Marcial1234:
Part of #247 
This will makes our protobuf compile to the latest version before compiling the entire tool. Protobuf syntax errors will appear upon `make`. 

In a Future PR, once proto definitions stabilize, simple definition and extraction tests will be done to all proto fields.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bazel-*
 .drone.sec.yml
 .idea/
 debug.test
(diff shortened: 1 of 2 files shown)


cassierecher:
Please add documentation that you must have the protobuf compiler with Go runtime installed, and `protoc` in your path.

PR #265, opened 2019-05-24
[design-review] New Rendering Process

Marcial1234:
**Note**: Do not review this PR's files, this is a mini-design review request. Actual code further down on this post. Thought of too many permutations to code them all in concrete files, but I provide the main modular pieces inline below. Ideas are post-PoC, as all are achievable. Will aid #247.

## Summary
I'd like to discuss the implementation of the future proto to md-and-html render. Given the possibility of future "quick fixes" and small feature additions of this repo, which of these processes (or any other) can serve as an optimal technical-debt minimizing default-architecture? Where do we draw the line on over-engineering?

### Rendering
```go
import "path/to/devrel_tutorial"
t := &devrel_tutorial.Tutorial{...}
...
t.Md()   // Outputs as Markdown
t.Html() // Outputs as HTML
```

### Adding new proto `message` type
1. Create new type in `tutorial.proto`
2. Implement `ProtoRenderer`'s `Html` and `Md` interface methods inside each extension's renderer file. Base them on another `Leaf` type element declaration if your new field doesn't include other subtypes.
3. Add new `html` and `md` templates for your type.
4. Add new `oneof` type mapping to `render_oneof.go`(link TBA)
5. _TBD_ once `repeated` approach is settled between template vs possibly-concurrent approaches.

### Modifying an existing type
- Modify template according to field change
- Follow steps above starting at #4 for adding new `oneof` and `repeated` fields.
---

### Proto fields standardization/predictability
Ideally, for the most generalizable code, the follow could be enforced:
- Name all `oneof` fields `content` (that's already the case).
- Name all `repeated` fields `children` => `repeated Type children = 888;`
  - This takes away some semantic context, but given we already have a type, having a generic field name makes generalizations later in code easier.

### Templating Idea
- We decouple the conditional templating logic and their rendering. Like so, the smallest writable units are proto elements. Some conditional logic will be maintained, but it'll not be as messy to read.
- We write one proto-type-template per extension and store them as files _without extensions_ to maintain a proto-message-type-to-template-name mapping, and they will be stored in an extension-based directory (HTML templates under `html`, etc).
- We register the templates per path easily registered with [`template.ParseGlob`](https://golang.org/pkg/text/template/#ParseGlob), and they'll be rendered using package extension-specific variables. 

#### Templating Code
```go
// in globals-renderer-pkg file
const (
  mdTmplsDir := "path/to/template/md/*",
  htmlTmplsDir := "path/to/template/html/*",
)
var (
  md   *template.Template
  html *template.Template
)

// "path/to/proto-pkg/render_md.go" file
md = template.Must(template.New("master").ParseGlob(mdTmplsDir))

// "path/to/proto-pkg/render_html.go" file
html = template.Must(template.New("master").ParseGlob(htmlTmplsDir))

// "path/to/template/html/Heading" file
<h{{.Level}}> {{- .Text -}} </h{{.Level}}>

// "path/to/template/md/Heading" file
##{{ repeatedHeading .Level }} {{.Text -}}
```
-----

### Proto Rendering
The current rendering logic, but returning strings would be as follows:
```go
func render(data interface{}, t *template.Template) string {
  var r string

  switch data.(type) {
  case *devrel_tutorial.Heading:
    r = executeHeadingTemplate(data, t)
    // all others ...
  }

  return r
}
...
func executeHeadingTemplate(data interface{}, t *template.Template) string {
  var w bytes.Buffer
  util.LogIfError(t.ExecuteTemplate(&w, "Heading", data))
  return w.String()
}
// global util pkg
func LogIfError(err error) {
  if err != nil {
    log.Fatal(err)
  }
}
```

#### Proto Rendering Ideas
- Generic rendering methods `executeTemplate("heading")` instead of 1:1 methods `executeHeadingTemplate` since we'd already have them registered as templates.

##### Proto definitions extension
- We extend the Proto's package definition by adding interface methods for all subtypes that will take care of the rendering without having to maintain an extra master switch mapping. Having this logic in its own packages although nice, posed medium type-headaches. See *PS* at the bottom.
- *nit*: Template logic could be stored relative to proto-package - although not feeling too strongly about this one. 

```go
// in globals-render-pkg file
type ProtoRenderer interface {
  Md() string
  Html() string
}
...
func executeTemplate(d interface{}, tName string, t *template.Template) string {
  var w bytes.Buffer
  utils.LogIfError(t.ExecuteTemplate(&w, tName, d))
  return w.String()
}

// in extension-specific-render-file
function (el *Heading) Html() {
  return executeTemplate(&el, "Heading", htmlTemplates)
}

// any file that imports proto pkg
element := &devrel_tutorial.Heading{
  Text:  "hello",
  Level: 2,
}
fmt.Printf(element.Html())
// per path/to/template/html/Heading above, output: <h2>hello</h2>
```

#### Still debating: Global buffer/writer
Similar reasoning for template reuse, although not feeling too strongly about it since it can limit future functionality.

- Using it makes the code slightly uglier, having to pass the buffer/io.Writer variable along all the time (non-proto rendering code works like this). This does not lend itself to concurrency in the future at first sight, since it's append-only.
- Not reusing it might reduce performance since we create a buffer to output strings for each rendering, but string returns could be easily concatenated concurrently with `strings.Join` for a more appropriate recursively-tree-traversal rendering.

#### Still debating: Rendering Overhead, where to execute `repeated` fields?
Rendering the `oneof` and `repeated` containing types (`BlockContent`, `Steps`, etc) pose a rendering overhead to the desired generic and modular templates design. `Step`s and `Tutorial` can stay continue to be rendered on a single template (else a new `int number` field would be needed on `Step` - although is it actually useful?). I am hesitant to follow this approach for the rest, but might do so for the sake of speed. 

**PS**: Other methods were attempted to maximize interface/API generalization, but did not come to fruition due to the limited support of built-in type-catch-alls for proto-messages, and although one such-type could be made, it's still non-trivial to generalize conversations as per the type differences in `repeated / []`, `oneof`, and leaf-type elements. Custom switch-based logic cannot be escaped for `oneof` types, but this is a necessary overhead and semantically useful.


Marcial1234:
Closing as not a real PR to not get notifications. Still open for comments on overall design.

Issue #99, opened 2018-10-04
Embed surveys inside the codelab


mrina-natarajan:
Quizzes and other forms of survey checkpoints are essential to know whether the students learn and absorb concepts and knowledge as designed in the training.

Right now, we can only point to a link for a google survey. It's better to make that testing experience also engaging by embedding it into the codelab directly. 

We want to embed a google form itself inside the codelab for proactive feedback.

PR #268, opened 2019-05-26
add support for embeddable iframes

marcacohen:
- reuse mechanism Nicholas Garnier established for embedding youtube videos
- set alt param to iframe src URL
- URL must be in a global whitelist of embeddable domains
- if not whitelisted, pass through as a normal embedded image


FORMAT-GUIDE.md
```diff
@@ -60,9 +60,16 @@ You can also use this to target specific events, for instance:  \
...
-     - Put a Youtube video link in the **Description** field of the Alt Text. in the format `https://www.yoututbe.com/watch?v=[video_ID]`
+     - Put a Youtube video link in the **Description** field of the Alt Text. in the format `https://www.youtube.com/watch?v=[video_ID]`
     > Specifying a start time is not supported at this time.
 
+1. Embedded Iframes
+
+    Iframes can be embedded by doing:
+     - Add an image in the document. The image can be a screenshot of the iframe for instance but it doesn't really matter since it won't be displayed but replaced by the embedded iframe.
+     - Add an "Alt Text" to the image by doing **Cmd+Opt+Y** or **Right click > "Alt Text..."**
+     - Put a full URL in the **Description** field of the Alt Text. in the format `https://www.domain.com/watch?foo=bar`
```
nicolasgarnier: Add a note that only httpS is supported and not http? :)

claat/parser/gdoc/parse.go
```diff
@@ -650,9 +650,27 @@ func list(ds *docState) types.Node {
 // image creates a new ImageNode out of hn, parsing its src attribute.
 // It returns nil if src is empty.
 // It may also return a YouTubeNode if alt property contains specific substring.
+// or an IframeNode if the alt property contains a URL other than youtube.
 func image(ds *docState) types.Node {
-	if strings.Contains(nodeAttr(ds.cur, "alt"), "youtube.com/watch") {
+	alt := nodeAttr(ds.cur, "alt")
+	if strings.Contains(alt, "youtube.com/watch") {
 		return youtube(ds)
+	} else if strings.Contains(alt, "https://") {
```
nicolasgarnier: Could we do "Starts With" rather than "Contains" otherwise a non-https URL that has "https://" in the URL (as a query param for instance) will be accepted :P)

nicolasgarnier: Ah it's fine since you double check afterwards in `func iframe`

claat/parser/gdoc/parse.go
```diff
@@ -650,9 +650,27 @@ func list(ds *docState) types.Node {
...
+		u, err := url.Parse(nodeAttr(ds.cur, "alt"))
+		if err != nil {
+			return nil
+		}
+		// For iframe, make sure URL ends in whitelisted domain.
+		ok := false
+		for _, domain := range types.IframeWhitelist {
```
nicolasgarnier: It would be cool to display some type of error or warning somehow (not sure how to easily do that) when the domain has not be whitelisted. For instance by adding some Warning text to the "alt" of the image.

If we could do something like `setNodeAttr(ds.cur, "alt", "The domain of the requested iframe <domain here> has not been whitelisted.")` woud work since we are adding the "alt" text to images.

nicolasgarnier:
Some nits :) Otherwise LGTM pending on fixing tests :)

FORMAT-GUIDE.md
marcacohen: good point, fixed.

claat/parser/gdoc/parse.go
marcacohen: good idea, fixed.

marcacohen:
@nicolasgarnier test cases and review comments fixed, PTAL

claat/parser/gdoc/parse.go
nicolasgarnier: Cool! Could you also print a warning? it will make it easier to debug and maybe easier for external people that runs claat standalone. :)

nicolasgarnier:
Just one last nit about printing a warning in the console about the domain whitelist  :)

claat/parser/gdoc/parse.go
marcacohen: Great idea, added, and thanks for the presumptive approval. Will merge after build passes travis.

marcacohen:
This closes #99.

PR #272, opened 2019-05-29
[proto-render-setup Part 3]: Base Testing logic, ideal TODOs

Marcial1234:
Part of #247, all tests pass depending on #254 definitions. Ready for nits!

End of setup batch. Will dump element-by-element rendering + template + proto changes for the next stacked PRs batch.

### DIFF
--- a/third_party/render_html.go
+++ b/third_party/render_html.go
@@ -1,19 +1,20 @@
 package devrel_tutorial
 
 import (
-	"strings"
 	"text/template"
 )
 
 func init() {
-	funcMap := template.FuncMap{
-		"safeHTML": func(s string) htmlTemplate.HTML {
-			return template.HTML(s)
-		},
-	}
-	html = template.Must(template.New("master").Funcs(funcMap).ParseGlob(htmlTmplsDir))
+	html = template.Must(template.New("master").ParseGlob(htmlTmplsDir))
 }
 
+// TODO if possible:
+//     Template names are named after their struct type,
+//     use `reflect.TypeOf(el).Elem().Name()`
+//     as a generic caller once a catch-all type, or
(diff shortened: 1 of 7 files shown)


third_party/render_html.go
```diff
@@ -1,15 +1,20 @@
...
 )
 
 func init() {
-	funcMap := template.FuncMap{
-		"safeHTML": func(s string) htmlTemplate.HTML {
-			return template.HTML(s)
-		},
-	}
-	html = template.Must(template.New("master").Funcs(funcMap).ParseGlob(htmlTmplsDir))
+	html = template.Must(template.New("master").ParseGlob(htmlTmplsDir))
```
shawnbuso: Is this supposed to be undoing stuff that was in PR 2? If so should we just remove it from PR 2?

third_party/render_md_test.go
```diff
@@ -0,0 +1,27 @@
+package devrel_tutorial
+
+import (
+	"testing"
+	"github.com/googlecodelabs/tools/claat/util"
+)
+
+func TestRenderHeadingMd(t *testing.T) {
+	h2 := &Heading{
+    Level: 2,
```
shawnbuso: Fix indentation here and below

Marcial1234: aha! good catch! I'm using Sublime text with same size tabs as two-spaces. Didn't catch this. This repo uses tabs for default spacing, for some reason.

third_party/render_html.go
Marcial1234: I found a better way to do this, so it's not an undo. But the change definitely needs to be on Part 2. Will clean that one in general a bit.

Marcial1234:
Hold on further reviewing until Design Doc is approved, closing for now

PR #273, opened 2019-05-29
[proto-render-templating Part 1] 'StylizedText'. Allows for simultaneous italics and boldening

Marcial1234:
Part of #247

In a future PR, `Links` will also be allowed to be of simultaneously bold and italics.
Missing from `InlineContent`: yet-to-be-pushed `InlineCode` proto-type.

### DIFF
--- a/third_party/tutorial.proto
+++ b/third_party/tutorial.proto
@@ -59,13 +59,21 @@ message Link {
 
 message InlineContent {
   oneof content {
-    string text = 1;
+    StylizedText text = 6;
     string strong_text = 2;
     string emphasized_text = 3;
     string inline_code_text = 4;
     Link link = 5;
   }
+
+  reserved 1;
 };
+
+message StylizedText {
+  string text = 1;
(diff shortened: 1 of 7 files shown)


third_party/tutorial.proto
```diff
@@ -59,14 +59,20 @@ message Link {
...
-    string text = 1;
-    string strong_text = 2;
-    string emphasized_text = 3;
-    string inline_code_text = 4;
+    StylizedText text = 6;
     Link link = 5;
   }
+
+  reserved 1, 2, 3, 4;
+  reserved "strong_text", "emphasized_text", "inline_code_text";
```
shawnbuso: I'm not sure if re-using field names but changing the index works (e.g. re-using "text" but changing the index from 1 to 6). Cassie?

Marcial1234: I'm not sure either. Compiles, but I don't know if that makes any difference. @cassierecher

PR #280, opened 2019-06-13
Add beginnings of a more modular fetch system

cassierecher:
Not yet totally completed or integrated into the existing code. Will continue to work.

### DIFF
--- a/claat/fetch/filesystem/filesystem_test.go
+++ b/claat/fetch/filesystem/filesystem_test.go
@@ -1,0 +1,53 @@
+package filesystem
+
+import (
+	"io/ioutil"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestNewFileSystemFetcher(t *testing.T) {
+	p := "this/is/a/file/path"
+	fsf := NewFileSystemFetcher(p)
+
+	if fsf.resPath != p {
+		t.Errorf("NewFileSystemFetcher(%v).resPath = %v, want %v", p, fsf.resPath, p)
+	}
+}
+
+func TestFetch(t *testing.T) {
+	// Make temporary file for testing purposes.
+	contents := []byte("file contents!")
+	f, err := ioutil.TempFile("", "filesystem_test_file")
+	if err != nil {
+		t.Errorf("error creating temp file: %s", err)
+	}
+	fname := f.Name()
+	defer os.Remove(fname)
+
+	// Write some bytes to the file.
+	_, err = f.Write(contents)
+	if err != nil {
+		// Be sure to close the file if we're going to stop here.
+		f.Close()
+		t.Errorf("error writing to temp file: %s", err)
+	}
+
+	f.Close()
+
+	fsf := NewFileSystemFetcher(fname)
+	r, err := fsf.Fetch()
+	if err != nil {
+		t.Errorf("Fetch() = got err %v, want nil", err)
+	}
+
+	// Get the bytes out of the reader and compare.
+	res, err := ioutil.ReadAll(r)
+	if err != nil {
+		t.Errorf("error reading from Fetch() result: %s", err)
+	}
+	if !reflect.DeepEqual(res, contents) {
+		t.Errorf("Fetch() reader got %v, want %v", res, contents)
+	}
+}
--- a/claat/fetch/filesystem/filesystem_test.go
+++ b/claat/fetch/filesystem/filesystem_test.go
@@ -28,13 +28,10 @@ func TestFetch(t *testing.T) {
 
 	// Write some bytes to the file.
 	_, err = f.Write(contents)
+	defer f.Close()
 	if err != nil {
-		// Be sure to close the file if we're going to stop here.
-		f.Close()
 		t.Errorf("error writing to temp file: %s", err)
 	}
-
-	f.Close()
 
 	fsf := NewFileSystemFetcher(fname)
 	r, err := fsf.Fetch()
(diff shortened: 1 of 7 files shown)


claat/fetch/remotehttp/remotehttp.go
```diff
@@ -0,0 +1,46 @@
...
+	return rhf
+}
+
+// Fetch fetches the resource.
+func (rhf RemoteHTTPFetcher) Fetch() (io.Reader, error) {
+	res, err := rhf.c.Get(rhf.url)
+	if err != nil || res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error making HTTP request: status code %v and err %s", res.StatusCode, err)
+	}
+	defer res.Body.Close()
```
Marcial1234: any benefit from deferring here vs closing after we're done reading `res.Body` below?

claat/fetch/remotehttp/remotehttp_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		{
+			in: input{
+				statusCode: http.StatusNotFound,
+			},
+			out: output{},
+		},
+	}
+	for _, tc := range tests {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
Marcial1234: nit: line longer than 80 characters

claat/fetch/inprocess/inprocess.go
```diff
@@ -0,0 +1,22 @@
+package inprocess
+
+import "io"
+
+// InProcessFetcher implements fetch.Fetcher. It retrieves resources from an in-process source via io.Reader.
```
Marcial1234: nit: line longer than 80 characters

claat/fetch/remotehttp/remotehttp.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return rhf
+}
+
+// Fetch fetches the resource.
+func (rhf RemoteHTTPFetcher) Fetch() (io.Reader, error) {
+	res, err := rhf.c.Get(rhf.url)
+	if err != nil || res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error making HTTP request: status code %v and err %s", res.StatusCode, err)
```
Marcial1234: nit: line longer than 80 characters

claat/fetch/remotehttp/remotehttp_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+)
+
+func TestFetch(t *testing.T) {
+	type input struct {
```
Marcial1234: nit: Benefit of having these as named types vs anonymous `structs`? or their inner types since the structure is simple enough?

maybe something like this:
```go
tests := []struct {
    inStatusCode  int
    wantError     bool
}{
```

`out.ok` seems redundant since it means `true`. 

This format seems to fit more complicated tests with various inputs and outputs args.

claat/fetch/filesystem/filesystem_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Errorf("error creating temp file: %s", err)
+	}
+	fname := f.Name()
+	defer os.Remove(fname)
+
+	// Write some bytes to the file.
+	_, err = f.Write(contents)
+	if err != nil {
+		// Be sure to close the file if we're going to stop here.
+		f.Close()
```
Marcial1234: why not have this/`f.Close()` as a `defer` after `defer os.Remove` above if it is a wanted behavior no matter what happens? (`defer` order seems to be LIFO). then you'd get rid of the double mention?

claat/fetch/remotehttp/remotehttp_test.go
cassierecher: 80 char limit isn't typically enforced in Go files

claat/fetch/remotehttp/remotehttp.go
cassierecher: defer is guaranteed to run if we exit early. We don't have the added overhead of managing the location of calling Close if this code changes.

claat/fetch/remotehttp/remotehttp.go
cassierecher: 80 char limit isn't typically enforced in Go files

claat/fetch/inprocess/inprocess.go
cassierecher: 80 char limit isn't typically enforced in Go files

claat/fetch/filesystem/filesystem_test.go
cassierecher: Good idea

claat/fetch/remotehttp/remotehttp_test.go
cassierecher: The in/out structure is a common pattern: http://go/gh/golang/go/wiki/TableDrivenTests

Marcial1234: right. I mainly saw it as an overkill for this case (these are the main things I think about... welp)

claat/fetch/remotehttp/remotehttp.go
Marcial1234: "We don't have the added overhead of managing ... if this code changes" I like it. Didn't think of it that way before.

claat/fetch/remotehttp/remotehttp_test.go
cassierecher: I'd prefer "overkill" to using a nonstandard pattern.

PR #281, opened 2019-06-17
[reviewed-proto-renderer-setup]: Generic Renderer with sample template namespace tests

Marcial1234:
A restart of #247 

Base logic for future protobuf-based renderer. On this one we use pure reflection to route types to their templates, but this will changes to a type-switch on a future PR.

### DIFF
--- a/claat/proto-renderer/generic_renderer.go
+++ b/claat/proto-renderer/generic_renderer.go
@@ -14,7 +14,7 @@ func ExecuteTemplate(el interface{}, t *template.Template) string {
 	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
 	if e != nil {
 		// This method outputs directly to templates. Panicking to surfance errors
-		//   since we cannot handle multiple returns in templates
+		// since we cannot handle multiple returns in templates.
 		// Panics will be gracefully handled by a wrapper function
 		panic(fmt.Sprintf("Templating panic: %s\n", e))
 	}
--- a/claat/proto-renderer/generic_renderer.go
+++ b/claat/proto-renderer/generic_renderer.go
@@ -1,0 +1,19 @@
+package genrenderer
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"text/template"
+)
+
+// ExecuteTemplate returns the evaluated template per passed templating
+// namespace, based on the passed Codelab proto element type
+func ExecuteTemplate(el interface{}, t *template.Template) string {
+	var w bytes.Buffer
+	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
+	if e != nil {
+		panic(fmt.Sprintf("Templating err: %s\n", e))
+	}
+	return w.String()
+}
--- a/claat/proto-renderer/generic_renderer.go
+++ b/claat/proto-renderer/generic_renderer.go
@@ -13,7 +13,10 @@ func ExecuteTemplate(el interface{}, t *template.Template) string {
 	var w bytes.Buffer
 	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
 	if e != nil {
-		panic(fmt.Sprintf("Templating err: %s\n", e))
+		// This method outputs directly to templates. Panicking to surfance errors
+		//   since we cannot handle multiple returns in templates
+		// Panics will be gracefully handled by a wrapper function
+		panic(fmt.Sprintf("Templating panic: %s\n", e))
 	}
 	return w.String()
 }
--- a/claat/proto-renderer/generic_renderer_test.go
+++ b/claat/proto-renderer/generic_renderer_test.go
@@ -8,9 +8,9 @@ import (
 )
 
 type encapsulatedTest struct {
-	In      sampleProtoTemplate
-	Out     string
-	WantErr bool
+	in  sampleProtoTemplate
+	out string
+	ok  bool
 }
 
 type sampleProtoTemplate struct {
@@ -19,25 +19,6 @@ type sampleProtoTemplate struct {
 
 func newSampleProtoTemplate(el interface{}) sampleProtoTemplate {
 	return sampleProtoTemplate{Value: el}
-}
-
-func runEncapsulatedTest(test encapsulatedTest, tmpl *template.Template, t *testing.T) {
-	// Check wheather template failed to render by checking if there was a panic
-	defer func(test encapsulatedTest) {
-		err := recover()
-		if err != nil && !test.WantErr {
-			t.Errorf("For: %#v\n\t%s", test, err)
-		}
-
-		if err == nil && test.WantErr {
-			t.Errorf("Error did not occur for: %#v", test)
-		}
-	}(test)
-
-	tmplOut := ExecuteTemplate(test.In, tmpl)
-	if test.Out != tmplOut && !test.WantErr {
-		t.Errorf("Expecting:\n\t'%s', but got \n\t'%s'", test.Out, test.In)
-	}
 }
 
 func TestExecuteTemplate(t *testing.T) {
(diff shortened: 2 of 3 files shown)


claat/proto-renderer/generic_renderer.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+	"reflect"
+	"text/template"
+)
+
+// ExecuteTemplate returns the evaluated template per passed templating
+// namespace, based on the passed Codelab proto element type
+func ExecuteTemplate(el interface{}, t *template.Template) string {
+	var w bytes.Buffer
+	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
```
Marcial1234: Per PR description, this will change to a function-based type-switch on a later PR. This is the base functionality per this setup PR.

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+type sampleProtoTemplate struct {
+	Value interface{}
+}
+
+func newSampleProtoTemplate(el interface{}) sampleProtoTemplate {
+	return sampleProtoTemplate{Value: el}
+}
+
+func runEncapsulatedTest(test encapsulatedTest, tmpl *template.Template, t *testing.T) {
```
Marcial1234: Need to encapsulate the scope of possible panics, else tests won't be executed past the first panic on a loop.

claat/proto-renderer/generic_renderer.go
Marcial1234: ideally this will be a proto, but the definitions of leaf-types need to be redefined first. Showcasing generality here.

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"go/build"
+	"path/filepath"
+	"testing"
+	"text/template"
+)
+
+type encapsulatedTest struct {
+	In      sampleProtoTemplate
+	Out     string
+	WantErr bool
```
Marcial1234: any clearer naming? `ShouldPanic`/`ShouldErr`?

claat/proto-renderer/generic_renderer.go
```diff
@@ -0,0 +1,19 @@
...
+	"bytes"
+	"fmt"
+	"reflect"
+	"text/template"
+)
+
+// ExecuteTemplate returns the evaluated template per passed templating
+// namespace, based on the passed Codelab proto element type
+func ExecuteTemplate(el interface{}, t *template.Template) string {
+	var w bytes.Buffer
```
cassierecher: bytes.NewBuffer

claat/proto-renderer/generic_renderer.go
```diff
@@ -0,0 +1,19 @@
...
+	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
+	if e != nil {
+		panic(fmt.Sprintf("Templating err: %s\n", e))
```
cassierecher: Put a comment about why this and not returning an error

claat/proto-renderer/generic_renderer_test.go
cassierecher: Conventionally these should be "in", "out" and "ok". Non-public and without want/should.

claat/proto-renderer/generic_renderer_test.go
cassierecher: Have you checked to make sure this actually works? I have a suspicion there might be scoping issues associated with performing test ops in deferred code.

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,59 @@
...
+		"returnInt": func(i int) int { return i },
+	}
+	tmpl := template.Must(template.New("dummy").Funcs(funcMap).ParseGlob(tmplsAbsDir))
+
+	tests := []encapsulatedTest{
+		{newSampleProtoTemplate(3), "3", false},
+		{newSampleProtoTemplate("not-valid"), "", true},
+	}
+
+	for _, test := range tests {
```
cassierecher: Conventionally `_, tc`

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -24,13 +24,13 @@ func newSampleProtoTemplate(el interface{}) sampleProtoTemplate {
...
-			t.Errorf("False positive panic for: %+v", test)
+		err := recover()
+		if err != nil && !test.WantErr {
+			t.Errorf("For: %#v\n\t%s", test, err)
 		}
 
-		if r == nil && test.WantErr {
-			t.Errorf("False negative panic for: %+v", test)
+		if err == nil && test.WantErr {
+			t.Errorf("Error did not occur for: %#v", test)
```
cassierecher: Call it panic if it's a panic we're watching for

claat/proto-renderer/generic_renderer_test.go
Marcial1234: so copying your fetching PR's format? sg

claat/proto-renderer/generic_renderer_test.go
Marcial1234: what do you mean by "actually works"? the tests pass and flipping the `WantErr` booleans cause tests to break - as expected.

claat/proto-renderer/generic_renderer.go
Marcial1234: per https://golang.org/pkg/bytes/#NewBuffer, this requires an initializing parameter, which we don't really use or need for this method.

I got this "formatting style" per this conference talk (before: https://youtu.be/WkzGQ-Sm_50?t=439, after: https://youtu.be/WkzGQ-Sm_50?t=639)

claat/proto-renderer/generic_renderer_test.go
Marcial1234: update: verified having an inline func instead of a helper still works. helper functions seem fine or anything needs to change? at least their order within the file?

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+type sampleProtoTemplate struct {
+	Value interface{}
+}
+
+func newSampleProtoTemplate(el interface{}) sampleProtoTemplate {
+	return sampleProtoTemplate{Value: el}
+}
+
+func TestExecuteTemplate(t *testing.T) {
```
Marcial1234: Changed function order, helper now on bottom

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{newSampleProtoTemplate(nil), "", true},
+		{newSampleProtoTemplate("not-valid"), "", true},
+	}
+
+	for _, tc := range tests {
+		runEncapsulatedTest(tc, tmpl, t)
+	}
+}
+
+func runEncapsulatedTest(test encapsulatedTest, tmpl *template.Template, t *testing.T) {
```
Marcial1234: This can be an inline/nested func instead of a helper, and still works. helper functions seem fine?

Marcial1234:
I guess broken Travis solves the uncertainty of whether tests are being runned. Will TAL tomorrow

claat/proto-renderer/generic_renderer.go
```diff
@@ -0,0 +1,22 @@
...
+
+// ExecuteTemplate returns the evaluated template per passed templating
+// namespace, based on the passed Codelab proto element type
+func ExecuteTemplate(el interface{}, t *template.Template) string {
+	var w bytes.Buffer
+	e := t.ExecuteTemplate(&w, reflect.TypeOf(el).Name(), el)
+	if e != nil {
+		// This method outputs directly to templates. Panicking to surfance errors
+		//   since we cannot handle multiple returns in templates
+		// Panics will be gracefully handled by a wrapper function
```
shawnbuso: nit: Fix spacing and add periods to ends of comments.

claat/proto-renderer/generic_renderer.go
shawnbuso: If this is expected to change eventually, is there a reason we're not just changing it now?

Marcial1234: - Proto definitions need to be redefined and softly-agreed on
- Smaller reviews

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -54,7 +35,26 @@ func TestExecuteTemplate(t *testing.T) {
...
+		}
+
+		if err == nil && !test.ok {
+			t.Errorf("\nPanic did not occur for:\n\t%#v\n\t(false positive)", test)
+		}
+	}(test)
+
+	tmplOut := ExecuteTemplate(test.in, tmpl)
+	if test.out != tmplOut {
+		t.Errorf("Expecting:\n\t'%s'\nBut got:\n\t'%s'", test.out, tmplOut)
```
cassierecher: Try output format like:

ExecuteTemplate(%v, %v) = %s, want %s

It notes the input args and prominently names the problematic function as well.

claat/proto-renderer/generic_renderer_test.go
```diff
@@ -0,0 +1,60 @@
...
```
cassierecher: Try output format like:

ExecuteTemplate(%v, %v) = %s, want %s

It notes the input args and prominently names the problematic function as well.

claat/proto-renderer/generic_renderer_test.go
Marcial1234: > names the problematic function as well

I thought the test function name (`TestExecuteTemplate`) already defines that, but ok. Right now outputs look as following:

image

having `%#v` vs plain `%v` shows the inner type values, which can be helpful

Marcial1234: Now:
image

claat/proto-renderer/generic_renderer_test.go
cassierecher: Inline is fine.

PR #288, opened 2019-06-21
[reviewed-proto-renderer Part 1.1]: Base Leaf type rendering: StylizedText + Base testing functions

Marcial1234:
Part of #247 

Start of HTML Template rendering.
*Lots* of helper functions for testing... hopefully not overdoing it.

PR #290, opened 2019-06-25
[reviewed-proto-renderer Part 1.2]: Leaf type: InlineCode

Marcial1234:
Part of #247

In preparation for the `repeated InlineContent` field `TextBlock`, adding `InlineCode`.


claat/proto-renderer/html/templates-tests/inlinecode_test.go
```diff
@@ -0,0 +1,54 @@
...
+			true,
+		},
+		{
+			testingUtils.NewInlineCode("{ Αα Ββ Γγ Δδ Εε Ϝϝ Ζζ Ηη Θθ Ιι Κκ Λλ Μμ Νν Ξξ Οο Ππ Ρρ Σσς Ττ Υυ Φφ Χχ Ψψ Ωω }"),
+			"<code>{ Αα Ββ Γγ Δδ Εε Ϝϝ Ζζ Ηη Θθ Ιι Κκ Λλ Μμ Νν Ξξ Οο Ππ Ρρ Σσς Ττ Υυ Φφ Χχ Ψψ Ωω }</code>",
+			true,
+		},
+	}
+	testingUtils.CanonicalRenderingTestBatch(html.Render, tests, t)
+}
```
Marcial1234: Error here will look the name per https://github.com/googlecodelabs/tools/pull/288#discussion_r296940674, changed accordingly for their input/output

third_party/tutorial.proto
```diff
@@ -60,12 +60,12 @@ message Link {
 message InlineContent {
   oneof content {
     StylizedText text = 1;
-    string inline_code_text = 4;
+    InlineCode code = 6;
     Link link = 5;
```
shawnbuso: Keep these in index order, so code (6) should go after link (5).

claat/proto-renderer/html/templates-tests/inlinecode_test.go
cassierecher: Let's add the field names here too

PR #291, opened 2019-06-25
[reviewed-proto-renderer Part 2]: Base Oneof type rendering: InlineContent

Marcial1234:
Part of #247

`InlineContent` with Base Oneof rendering and testing process
Tests missing for `Link`, which will come in 2 PRs from now (#294)


claat/proto-renderer/oneof_utils.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// RenderOneof returns the underyling, rendered, passed oneof element
+func RenderOneof(el interface{}, t *template.Template) string {
+  // Recursive redering happens here
+  return ExecuteTemplate(AssertOneof(el), t)
+}
+
+// AssertOneof turns a generic oneof proto into its underlying typed-proto
+func AssertOneof(el interface{}) (underlyingType interface{}) {
```
shawnbuso: I feel like strong typing, like Go provides, should eliminate the need for this. We're basically just saying "what type is this thing?", no?

claat/proto-renderer/testing-utils/general.go
```diff
@@ -8,31 +8,32 @@ import (
...
 	"github.com/googlecodelabs/tools/third_party"
 )
 
 // unsupportedType is a dummy type used to showcase the failures of rendering
 // non-proto custom types since we take in "any" type as rendering input.
 type UnsupportedType struct{}
 
-// Simple proto constructor
+// NewDummyProto is a simple proto constructor
 func NewDummyProto(in string) *tutorial.StylizedText {
```
shawnbuso: If this only returns stylized text, should it be called "NewStylizedText"?

claat/proto-renderer/testing-utils/proto_constructors.go
```diff
@@ -39,3 +39,57 @@ func NewInlineCode(txt string) *tutorial.InlineCode {
 		Code: txt,
 	}
 }
+
```
shawnbuso: I'm not clear on why we need these - can't the code calling these methods just call the constructors directly, like these methods are doing?

claat/proto-renderer/oneof_utils.go
Marcial1234: Per the entire reflection discussion on the design doc:
- The only type in Go that can be used as generic/catch-up type is `interface{}`

What's not documented here per assumption of features of golang: The original type of an non-typed interface is lost once we send it down to another method. Therefore, this cannot be eliminated - since when we call `ExecuteTemplate`, `templateName` will fail to get the passed type.

Any suggestions on the function name welcomed. I can only think of `AssertUnderlyingOneof`, since that's the nomenclature I've been using.

claat/proto-renderer/testing-utils/general.go
Marcial1234: I wanted abstract out what type this really is, for readability distincion. No knowledge of actual protos is assumed on the tests (output-format base code) where this is used.

I'll make it not take in an argument and simply have "dummy" inside.

claat/proto-renderer/testing-utils/proto_constructors.go
Marcial1234: There's no constructors in Go since there are no classes. Only way to instantiate elements is by explicit definition. These are such constructors. I got this from Cassie's PR ... TBA

Marcial1234: this also makes tests easier to write and read, as per #292: 
https://github.com/googlecodelabs/tools/blob/e57932077015ed9a5f4ecda80e2eeec99d7793dc/claat/proto-renderer/html/templates-tests/textblock_test.go#L24

claat/proto-renderer/oneof_utils.go
Marcial1234: > should eliminate the need for this
If by `the need for this` do you mean "no need to have an extra function", and the intent is to merge both `RenderOneof` and `AssertOneof` into a single function, this was done for the following:
- Separation of concerns: type-assertions has nothing to with actual template rendering
- Later reusability: if later needed, `AssertOneof` can be used without further refactoring

claat/proto-renderer/oneof_utils.go
```diff
@@ -0,0 +1,48 @@
...
+  // debug-friendly panic
+  if underlyingType == nil {
+    panic(TypeNotSupported("AssertOneof", el))
+  }
+
+  return underlyingType
+}
+
+// InnerContentsInline asserts the underlying type of tutorial.InlineContent
+func InnerContentsInline(el *tutorial.InlineContent) (underlyingType interface{}) {
```
Marcial1234: in a later PR, I change this to `UnderlyingInlineContentType`

PR #292, opened 2019-06-26
[reviewed-proto-renderer Part 3]: Base Repeated type rendering: Paragraph

Marcial1234:
Part of #247 

`TextBlock` type introduced. Changed all protos that depend on it.

Non-reflection based iterative type-assertions conversion of a _catch-all_ `repeated field` protos into a [interface slice](https://github.com/golang/go/wiki/InterfaceSlice) introduced. This is foundation for all other `repeated` fields rendering.


claat/proto-renderer/repeated_utils.go
```diff
@@ -0,0 +1,54 @@
...
+		renderedEls[i] = ExecuteTemplate(contents[i], t)
+	}
+
+	return renderedEls
+}
+
+// AssertRepeated turns a generic proto slice into typed-slice that can be
+// interated over without reflection. Panics if the passed type is not
+// explicitly defined
+func AssertRepeated(el interface{}) (guaranteedProtoSlice []interface{}) {
```
shawnbuso: Same comment as other PR - this "assert" method is doing more than just asserting something.

Marcial1234: Per https://github.com/googlecodelabs/tools/pull/291#discussion_r297828437, suggestions on function names besides `AssertUnderlyingRepeated` welcomed.  This is a type-assertion wrapper

Marcial1234:
reserved proto names and fields per semantic changes

claat/proto-renderer/repeated_utils.go
Marcial1234: and in addition to the above comment, since the use of reflection was dropped, merging these two functions will result in an undable mess. Rather have a small, easy to read logic + a smaller, decoupled mess.

only downside here is that, if wanted, future concurrency will need to implemented in two places

shawnbuso: I guess what I'm trying to say is conventionally methods called "assert" do one of two things:

a) nothing (e.g. assertTrue(true) will do nothing)
b) raise an exception (e.g. assertTrue(false) will throw an error)

In this case we have an AssertX method that takes in a value and then gives you a different value in return. Conventional assert methods don't do that.

Marcial1234: Oh I see. Yeah coming from a testing-framework background this makes sense. I originally named these `TypeAssertX` but I thought "type" was redundant.

This method are doing the same functionality as your point #2, `AssertType(notType)` => `panic` (Golang's hard exceptions). I'll do research to see if point #1 _actually_ doesn't return anything.

At any rate, "Type asserts" are very much a thing (link TBA, check comment in a bit).

Marcial1234: type assert link included

claat/proto-renderer/html/templates/StylizedText
```diff
@@ -1,7 +1,7 @@
 {{ if .IsStrong -}} <strong> {{- end -}}
```
Marcial1234: nits: getting rid of over-trimming (what the `-` characters do)

PR #297, opened 2019-06-30
[reviewed-proto/renderer Part 6]: List: new type and dummy tests

Marcial1234:
Part of #247


claat/proto-renderer/html/render.go
```diff
@@ -54,3 +57,27 @@ func renderOneof(contents interface{}) string {
...
+		return "ul"
+	case "ORDERED":
+		return "ol"
+	default:
+		return "unknown-list-variety"
+	}
+}
+
+// listFormattingClass maps 'ListStyle' enums to their CSS classes
+func listFormattingClass(s tutorial.List_ListStyle) string {
```
shawnbuso: Can't we just use strings.ToLower(s.String())?

claat/proto-renderer/html/render.go
```diff
@@ -54,3 +57,27 @@ func renderOneof(contents interface{}) string {
 func renderRepeated(contents interface{}) []string {
 	return genrenderer.RenderRepeated(contents, tmplNmspc)
 }
+
+// listVariertyToTag maps 'ListVariety' enums to their HTML tags
+func listVariertyToTag(v tutorial.List_ListVariety) string {
```
cassierecher: Variety*

claat/proto-renderer/html/render.go
Marcial1234: not for the case of unknowns/variable not set, no. We can use the built-in enum-to-string proto function, but you'd have to be oke with with having unused CSS classes on the HTML output.

shawnbuso: Doesn't the fact that we're using an enum guarantee that we won't get unknowns? And is variable not set possible? I thought enum values got assigned defaults.

Marcial1234: discussed offline. Going with the following:

```go
v := strings.ToLower(s.String())
if v == "Unknown_...": 
  return ""

return v
```

PR #299, opened 2019-07-01
 [reviewed-proto-renderer Part 7]: Image: new type and dummy tests

Marcial1234:
Part of #247 

Next PR: `ImageBlock`


claat/proto/renderer/html/templates-tests/link_test.go
```diff
@@ -10,17 +10,15 @@ import (
 )
 
 func TestRenderLinkTemplate(t *testing.T) {
```
Marcial1234: colocation nit

PR #304, opened 2019-07-08
Implement pass_metadata flag

### LINKED ISSUE #283: Generic metadata fields
In order for codelab.json to contain metadata from the table in the source document, the metadata must be explicitly handled by the parsing code. There should be a way to specify new metadata fields that make it to codelab.json without code changes.

cassierecher:
Implements a new command line flag, pass_metadata, that allows you to specify additional metadata fields that will be copied to codelab.json from the codelab source document without transforming them. This lets you create new simple metadata fields without modifying the source code.

This resolves #283

### DIFF
--- a/claat/cmd/export.go
+++ b/claat/cmd/export.go
@@ -41,6 +41,8 @@ type CmdExportOptions struct {
 	GlobalGA string
 	// Output is the output directory, or "-" for stdout.
 	Output string
+	// PassMetadata are the extra metadata fields to pass along.
+	PassMetadata map[string]bool
 	// Prefix is a URL prefix to prepend when using HTML format.
 	Prefix string
 	// Srcs is the sources to export codelabs from.
@@ -91,7 +93,7 @@ func CmdExport(opts CmdExportOptions) int {
 // nothing is stored on disk and the only output, codelab formatted content,
 // is printed to stdout.
 func exportCodelab(src string, opts CmdExportOptions) (*types.Meta, error) {
-	clab, err := slurpCodelab(src, opts.AuthToken)
+	clab, err := slurpCodelab(src, opts.AuthToken, opts.PassMetadata)
 	if err != nil {
 		return nil, err
 	}
--- a/claat/cmd/fetch.go
+++ b/claat/cmd/fetch.go
@@ -70,7 +70,6 @@ type codelab struct {
 // The function will also fetch and parse fragments included
 // with types.ImportNode.
 func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab, error) {
-	// TODO use passMetadata
 	res, err := fetch(src, authToken)
 	if err != nil {
 		return nil, err
--- a/claat/cmd/fetch.go
+++ b/claat/cmd/fetch.go
@@ -64,11 +64,13 @@ type codelab struct {
 }
 
 // slurpCodelab retrieves and parses codelab source.
+// It takes the source, plus an auth token and a set of extra metadata to pass along.
 // It returns parsed codelab and its source type.
 //
 // The function will also fetch and parse fragments included
 // with types.ImportNode.
-func slurpCodelab(src, authToken string) (*codelab, error) {
+func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab, error) {
+	// TODO use passMetadata
 	res, err := fetch(src, authToken)
 	if err != nil {
 		return nil, err
--- a/claat/cmd/fetch.go
+++ b/claat/cmd/fetch.go
@@ -76,7 +76,9 @@ func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab
 		return nil, err
 	}
 	defer res.body.Close()
-	clab, err := parser.Parse(string(res.typ), res.body, parser.Options{})
+	clab, err := parser.Parse(string(res.typ), res.body, parser.Options{
+		PassMetadata: passMetadata,
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/claat/cmd/fetch.go
+++ b/claat/cmd/fetch.go
@@ -76,7 +76,7 @@ func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab
 		return nil, err
 	}
 	defer res.body.Close()
-	clab, err := parser.Parse(string(res.typ), res.body)
+	clab, err := parser.Parse(string(res.typ), res.body, parser.Options{})
 	if err != nil {
 		return nil, err
 	}
--- a/claat/cmd/fetch.go
+++ b/claat/cmd/fetch.go
@@ -75,9 +75,11 @@ func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab
 		return nil, err
 	}
 	defer res.body.Close()
-	clab, err := parser.Parse(string(res.typ), res.body, parser.Options{
-		PassMetadata: passMetadata,
-	})
+
+	opts := *parser.NewOptions()
+	opts.PassMetadata = passMetadata
+
+	clab, err := parser.Parse(string(res.typ), res.body, opts)
 	if err != nil {
 		return nil, err
 	}
--- a/claat/main.go
+++ b/claat/main.go
@@ -70,9 +70,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	passMeta := make(map[string]bool)
+	passMetadata := make(map[string]bool)
 	for _, v := range strings.Split(*pass_metadata, ",") {
-		passMeta[v] = true
+		passMetadata[v] = true
 	}
 
 	exitCode := 0
@@ -84,7 +84,7 @@ func main() {
 			ExtraVars:    extraVars,
 			GlobalGA:     *globalGA,
 			Output:       *output,
-			PassMetadata: passMeta,
+			PassMetadata: passMetadata,
 			Prefix:       *prefix,
 			Srcs:         flag.Args(),
 			Tmplout:      *tmplout,
@@ -96,7 +96,7 @@ func main() {
 			AuthToken:    *authToken,
 			ExtraVars:    extraVars,
 			GlobalGA:     *globalGA,
-			PassMetadata: passMeta,
+			PassMetadata: passMetadata,
 			Prefix:       *prefix,
 		})
 	case "help":
--- a/claat/main.go
+++ b/claat/main.go
@@ -72,7 +72,7 @@ func main() {
 
 	pm := make(map[string]bool)
 	for _, v := range strings.Split(*passMetadata, ",") {
-		pm[v] = true
+		pm[strings.ToLower(v)] = true
 	}
 
 	exitCode := 0
--- a/claat/main.go
+++ b/claat/main.go
@@ -26,6 +26,7 @@ import (
 	"log"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/googlecodelabs/tools/claat/cmd"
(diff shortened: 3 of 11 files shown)


claat/cmd/export.go
```diff
@@ -41,6 +41,8 @@ type CmdExportOptions struct {
 	GlobalGA string
 	// Output is the output directory, or "-" for stdout.
 	Output string
+	// PassMetadata are the extra metadata fields to pass along.
+	PassMetadata map[string]bool
```
Marcial1234: I'd make this a custom type

claat/cmd/fetch.go
```diff
@@ -64,17 +64,21 @@ type codelab struct {
...
 // slurpCodelab retrieves and parses codelab source.
+// It takes the source, plus an auth token and a set of extra metadata to pass along.
 // It returns parsed codelab and its source type.
 //
 // The function will also fetch and parse fragments included
 // with types.ImportNode.
-func slurpCodelab(src, authToken string) (*codelab, error) {
+func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab, error) {
+	// TODO use passMetadata
 	res, err := fetch(src, authToken)
```
Marcial1234: Todo still needed?

claat/parser/gdoc/parse.go
```diff
@@ -110,23 +110,31 @@ const (
...
+	passMetadata map[string]bool // set of metadata fields to pass along.
 }
 
 type stackItem struct {
 	cur   *html.Node
 	flags stateFlag
 }
 
+func newDocState() *docState {
+	ds := new(docState)
```
Marcial1234: From what I've read the `new` keyword is discouraged in favor of struct declarations.

Any special rational for this here since it doesn't seem to be doing anything special?

claat/types/meta.go
```diff
@@ -64,6 +65,12 @@ type Codelab struct {
 	Steps []*Step
 }
 
+func NewCodelab() *Codelab {
+	clab := new(Codelab)
```
Marcial1234: Same comment as other `new` keyword

claat/parser/md/parse.go
```diff
@@ -127,6 +127,13 @@ type stackItem struct {
 	cur *html.Node
 }
 
+func newDocState() *docState {
```
Marcial1234: Method duplicate given different parser. Might be a good idea of having another pkg with all these constructors by themselves.

claat/parser/gdoc/parse.go
```diff
@@ -401,6 +409,10 @@ func metaTable(ds *docState) {
 			ds.clab.Feedback = s
 		case "analytics", "analytics account", "google analytics":
 			ds.clab.GA = s
+		default:
+			if _, ok := ds.passMetadata[fieldName]; ok {
```
Marcial1234: nit: no comment, unlike MD parser

claat/parser/gdoc/parse.go
Marcial1234: I'd also make this method take in a variadic/optional arg for `css` since it seems to be a common assignment right after.

claat/types/meta.go
```diff
@@ -64,6 +65,12 @@ type Codelab struct {
...
+	clab.Extra = make(map[string]string)
```
Marcial1234: Why `make` here?

claat/main.go
```diff
@@ -68,27 +70,34 @@ func main() {
 		os.Exit(1)
 	}
 
+	pm := make(map[string]bool)
```
Marcial1234:
```go
var pm map[string]bool
```

But also, make this whole diff into its own function. Maybe `parsePassMetadata`

claat/types/meta.go
cassierecher: Without it, the map is uninitialized and assignment to it fails.

claat/cmd/export.go
cassierecher: map[string]bool is a common enough pattern that I think aliasing it wouldn't help readability much

claat/cmd/fetch.go
cassierecher: Removed

claat/parser/gdoc/parse.go
cassierecher: Replaced with struct init

claat/parser/md/parse.go
cassierecher: Worth looking into the the future but IMO out of scope for this change.

claat/types/meta.go
cassierecher: Refactored

claat/parser/gdoc/parse.go
cassierecher: Added

claat/main.go
cassierecher: Done

claat/parser/md/parse.go
Marcial1234: If you find yourself making more duplicate constructors, then yeah this should be how is done. I agree this is small enough but the pattern might emerge eventually.

I did something similar for the proto constructors here https://github.com/googlecodelabs/tools/blob/dev/claat/proto/constructor/helpers.go

claat/parser/md/parse.go
```diff
@@ -127,6 +127,14 @@ type stackItem struct {
...
+	ds := &docState{
+		clab: types.NewCodelab(),
+	}
+
+	return ds
+}
+
```
Marcial1234: I still don't see the added value of this constructor. It's smelly imo.

This seems to be a wrapper to another constructor, and the result is proceded by more initialization assignments. Feels like either a more generic constructor should be made (not too readable past one variadic arg), or it should be left as a struct init.

And ofc, is also WET.

Marcial1234: Also nit: do not allocate extra variable. `return &...`

claat/types/meta.go
Marcial1234: Empty Map literal doesn't work? `map[string]string{}`

Per 

- https://stackoverflow.com/questions/16959992/creating-map-with-without-make
- https://stackoverflow.com/questions/31064688/which-is-the-nicer-way-to-initialize-a-map-in-golang

claat/cmd/fetch.go
```diff
@@ -64,17 +64,20 @@ type codelab struct {
...
-func slurpCodelab(src, authToken string) (*codelab, error) {
+func slurpCodelab(src, authToken string, passMetadata map[string]bool) (*codelab, error) {
 	res, err := fetch(src, authToken)
 	if err != nil {
 		return nil, err
 	}
 	defer res.body.Close()
-	clab, err := parser.Parse(string(res.typ), res.body)
+	clab, err := parser.Parse(string(res.typ), res.body, parser.Options{
+		PassMetadata: passMetadata,
```
Marcial1234: This multiline declaration looks like JavaScript. I'd make the `parser.Options` call into a constructor.

maybe-out-of-scope-or-too-nit: smelly params of inline conversion to a string + passing another struct variable as param ... I'd make into `parseParams` helper f(x) or something.

claat/parser/md/parse_test.go
```diff
@@ -152,7 +154,48 @@ feedback link: https://www.google.com
...
+	wantMeta := types.Meta{
+		Title:      title,
+		ID:         "zyxwvut",
+		Authors:    "john smith",
+		Summary:    "abcdefghij",
+		Categories: []string{"not", "really"},
+		Tags:       []string{"kiosk", "web"},
+		Feedback:   "https://www.google.com",
+		GA:         "12345",
+		Extra: map[string]string{
```
Marcial1234: nit: gofmt? weird spacing. Have seen this before, but don't recall why it happens... multiline declaration?

claat/parser/md/parse.go
Marcial1234: update: seem like you're already doing this with the `NewCodelab` constructor in `types/meta`

claat/parser/md/parse.go
cassierecher: A docState struct created by itself is not ready for use. The inner clab field requires init, so docState needs init. A New* function is the pattern for handling init, by producing an object that has all fields zeroed/inited so the caller is not required to do any manual init work on its own.

(handled nit)

claat/cmd/fetch.go
cassierecher: There's nothing inherently wrong with a multiline declaration. A NewOptions function is called for because the map needs to be zeroed.

claat/parser/md/parse_test.go
cassierecher: That is the output of gofmt.

claat/parser/md/parse.go
Marcial1234: > The inner clab field requires init, so docState needs init
correct, yet this is all this f(x) is doing, and in 2/3 times it's used it requires some other manual assignment, which makes me think this is an overkill/unnecessary-overhead as opposed to inline struct init (since optional args will be annoying to enforce).

this is more of a comment on the bad design of coupling of everything into a `docState`, but is fine to continue ~

claat/parser/parse.go
```diff
@@ -26,12 +26,23 @@ import (
...
 	// ParseFragment is similar to Parse except it doesn't parse codelab metadata.
 	ParseFragment(r io.Reader) ([]types.Node, error)
 }
 
+// Container for parsing options.
+type Options struct {
+	PassMetadata map[string]bool
+}
+
+func NewOptions() *Options {
```
Marcial1234: nit: make this take a variadic (read 'optional') arg that carries an non-dummy-initialized `PassMetadata`

claat/parser/md/parse_test.go
```diff
@@ -152,7 +154,48 @@ feedback link: https://www.google.com
...
+feedback link: https://www.google.com
+extrafieldone: aaaaa
+extrafieldtwo: bbbbb
+
+---
+`
+	content += ("# " + title)
+
+	c := mustParseCodelab(content, parser.Options{
+		PassMetadata: map[string]bool{
```
Marcial1234: make this use the `NewOptions` constructor

claat/types/meta.go
cassierecher: The last convention for `make` I heard was to use it whenever possible because a single keyword is simpler and makes the author's intention more explicit than creating an empty struct and taking the address of it. I don't have a particularly strong standing either way.

claat/parser/md/parse.go
cassierecher: It's expected to need to manipulate an object after initializing it. I don't see this as overhead, rather, as reducing overhead. Callers don't need to manually perform identical init steps everywhere, they just call newDocState and it's ready for whatever particular use they have in mind.

Marcial1234: seems fine then

claat/types/meta.go
Marcial1234: Yeah I don't see how it's explicit. `make` reminds me of C/++ of a dynamic-memory/heap allocation, and non-sized maps don't really fit that category since maps are meant to be mutable (they can't be labelled as `const`)

Feel free to resolve as you see fit.

claat/parser/md/parse_test.go
cassierecher: Done, plus in a few other places

claat/types/meta.go
cassierecher: Switched to empty map declaration. Cannot combine this with the above line because Extra comes from the embedded Meta struct.

claat/parser/parse.go
cassierecher: This becomes progressively harder to manage as the number of fields increases, and it doesn't make a marked difference over users assigning to PassMetadata directly. I think this is better off as an initializer rather than a constructor.

Marcial1234: sgtm

claat/main.go
```diff
@@ -101,6 +107,16 @@ func main() {
 	os.Exit(exitCode)
 }
 
+// parsePassMetadata parses metadata fields to parse that are not explicitly handled elsewhere.
+// It expects the fields to be passed in as a comma separated list (extraneous spaces are autoremoved), and returns a set of strings.
+func parsePassMetadata(passMeta string) map[string]bool {
+	fields := make(map[string]bool)
```
Marcial1234: nit: since opted for non-`make` inits, change to literal.
GTG

cassierecher: Done

Marcial1234:
All good to merge

PR #307, opened 2019-07-10
[reviewed-proto-renderer Part 10]: CodeBlock

Marcial1234:
Part of #247 

Syntax highlighting per language hints are never parsed. Added this as TODO on proto


claat/proto/renderer/html/templates/StylizedText.html
```diff
@@ -1,7 +1,7 @@
 {{ if .IsStrong -}} <strong> {{- end -}}
-{{- if .IsEmphasized -}} <em> {{- end -}}
+{{ if .IsEmphasized -}} <em> {{- end -}}
 
 {{- .Text | html | js -}}
```
Marcial1234: nit: unnecessary trims

third_party/tutorial.proto
```diff
@@ -123,9 +123,16 @@ message InfoBox {
 };
 
 message CodeBlock {
-  repeated InlineContent inline_contents = 1;
-  string code_location = 2;
-  string language_hint = 3;
+  // TODO: Use this instead of static highlighting
+  // currently being used to determine
```
cassierecher: I think a line got cut off here?

Marcial1234: as in?? I see it as I have it. I am forcibly line wrapping after some under 60 character limit so that the Diff on here don't line wrap since tabs are shows as ~6 spaces/chars.

Marcial1234: Bump

cassierecher: > Use this instead of static highlighting currently being used to determine

Determine what?

Marcial1234: Oh I see

PR #310, opened 2019-07-11
[reviewed-proto-renderer Part 12]: Survey & SurveyGroup

Marcial1234:
Part of #247

Design change: split our `Survey` feature into two protos since we can have one survey with many questions.

Added Copyright header to _new `.go` files_


claat/proto/constructor/helpers.go
```diff
@@ -140,6 +140,13 @@ func NewInlineContentTextStrongAndEmphasized(txt string) *tutorial.InlineContent
 	}
 }
 
+func NewHeading(level int32, contentSlice ...*tutorial.InlineContent) *tutorial.Heading {
```
Marcial1234: There seems to be a weird side effect of all the merges I've done that shows this as duplicate. Haven't been able to fix it, but this doesn't really add new code.

third_party/tutorial.proto
```diff
@@ -22,24 +22,24 @@ message Tutorial {
   Title title = 1;
   map<string, string> metadata = 2;
   repeated Step steps = 3;
-};
```
Marcial1234: removed all terminal `message` `;` since they're not needed

PR #314, opened 2019-07-17
Refactor fetch logic

cassierecher:


claat/fetch/filesystem/filesystem.go
```diff
@@ -26,17 +26,17 @@ type FileSystemFetcher struct {
 	resPath string
 }
 
-// NewFileSystemFetcher returns a new, initialized FileSystemFetcher.
+// New returns a new, initialized FileSystemFetcher.
 // The input string is the path to the file to read the resource from.
-func NewFileSystemFetcher(resPath string) FileSystemFetcher {
-	return FileSystemFetcher{
+func New(resPath string) *FileSystemFetcher {
```
NEricN: When using, this will look like `filesystem.New()`, which isn't immediately obvious it's a fetcher.

Discussed offline, but can you either rename this to filesystemfetcher.go or use the original name?

cassierecher: Done!

PR #318, opened 2019-07-22
Consistent non-sized Map initializations

Marcial1234:
Per https://github.com/googlecodelabs/tools/pull/304#discussion_r301816815, map initializations were agreed to not have the `make` keyword unless they included an initial length.

This keeps the codebase consistent. After this change, there will be only [3 instances](https://github.com/googlecodelabs/tools/search?q=%22make%28map%22&unscoped_q=%22make%28map%22) of sized maps initializations.


claat/cmd/auth.go
```diff
@@ -57,19 +57,15 @@ var (
...
-func init() {
-	clients = make(map[string]*http.Client)
-}
-
 // driveClient returns an HTTP client which knows how to perform authenticated
 // requests to Google Drive API.
 func driveClient(authToken string) (*http.Client, error) {
 	clientsMu.Lock()
 	defer clientsMu.Unlock()
-	provider := providerGoogle + authToken;
```
Marcial1234: This seems to be a carry-on from an old one-time change. nevertheless, ok to include imo.

PR #331, opened 2019-07-31
Create a "Fetcher" struct to encapsulate fetch logic.

cassierecher:


claat/fetch/drive/auth/auth.go
```diff
@@ -37,117 +35,107 @@ const (
...
+			Source: ts,
+			Base:   rt,
+		},
+	}, nil
 }
 
 // tokenSource creates a new oauth2.TokenSource backed by tokenRefresher,
 // using previously stored user credentials if available.
-// If authToken is given, we disregard the value of provider.
+// If authToken is not given, we use the Google provider.
```
NEricN: "not given" -> "not given in the constructor/new", since this itself doesn't take in parameters.

claat/fetch/fetch.go
```diff
@@ -13,8 +13,419 @@
...
+// or fetched from remote location.
+type resource struct {
+	typ  srcType       // source type
+	body io.ReadCloser // resource body
+	mod  time.Time     // last update of content
+}
+
+// codelab wraps types.Codelab, while adding source type
+// and modified timestamp fields.
+type codelab struct {
```
NEricN: Should this be exported since it's returned in the exported func SlurpCodelab below?

claat/fetch/fetch.go
```diff
@@ -13,8 +13,419 @@
...
+	}
+
+	imap := make(map[string]string, count)
+	var err error
+	for i := 0; i < count; i++ {
+		r := <-ch
+		imap[r.file] = r.url
+		if r.err != nil && err == nil {
+			// record first error
+			err = fmt.Errorf("%s => %s: %v", r.url, r.file, r.err)
```
NEricN: Would it be helpful to aggregate all the errors?

claat/fetch/fetch.go
```diff
@@ -13,8 +13,419 @@
...
+		b, err = ioutil.ReadFile(imgURL)
+		ext = filepath.Ext(imgURL)
+	} else {
+		b, err = f.slurpRemoteBytes(u.String(), 5)
+		if string(b[6:10]) == "JFIF" {
+			ext = ".jpeg"
+		} else if string(b[0:3]) == "GIF" {
+			ext = ".gif"
+		} else {
+			ext = ".png"
```
NEricN: Should it also confirm that it's really a PNG, and error out otherwise?

claat/fetch/fetch.go
```diff
@@ -13,8 +13,419 @@
...
+func (f *Fetcher) slurpRemoteBytes(url string, n int) ([]byte, error) {
+	res, err := retryGet(f.authHelper.DriveClient(), url, n)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	return ioutil.ReadAll(res.Body)
+}
+
+// retryGet tries to GET specified url up to n times.
```
NEricN: Can you specify it uses exponential backoff?

claat/cmd/export.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2016-2019 Google LLC. All Rights Reserved.
```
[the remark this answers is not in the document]
cassierecher: I'm making them simultaneous with the changes that reflect the updates.

claat/newfetch/fetch.go
```diff
@@ -1,32 +1,20 @@
...
-		t.Fatal(err)
-	}
-	tok, err := src.Token()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if tok.AccessToken != token {
-		t.Errorf("tok.AccessToken = %q; want %q", tok.AccessToken, token)
-	}
+type Fetcher interface {
```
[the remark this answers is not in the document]
cassierecher: Removed

claat/fetch/fetch.go
cassierecher: Out of scope for this PR I think. That's nice to have, but to the best of my recollection this hasn't been an issue in practice that I've heard of.

claat/fetch/fetch.go
cassierecher: There's not much of a reason for other code to create codelab structs. Or is this type of thing a general anti-idiom and it should be exported just for readability's sake? WDYT?

claat/fetch/fetch.go
cassierecher: I think this does what we want?

claat/fetch/fetch.go
NEricN: This looks to be one of those issues with people on both sides, so this should be fine then.

claat/fetch/fetch.go
```diff
@@ -171,17 +172,19 @@ func (f *Fetcher) SlurpImages(src, dir string, steps []*types.Step) (map[string]
...
-	var err error
+	errBuilder := strings.Builder{}
 	for i := 0; i < count; i++ {
 		r := <-ch
 		imap[r.file] = r.url
-		if r.err != nil && err == nil {
-			// record first error
-			err = fmt.Errorf("%s => %s: %v", r.url, r.file, r.err)
+		if r.err != nil {
+			errBuilder.WriteString(fmt.Sprintf("%s => %s: %v", r.url, r.file, r.err))
```
NEricN: I think these append them [without newlines](https://golang.org/src/strings/builder.go). Can you add newlines or some separator?

PR #338, opened 2019-08-07
Remove non-hermetic test cases from fetch_test.

cassierecher:
These tests were dependent on there being a token already written to
disk by a previous run of the binary. The functionality will be tested
in the future, along with a bit of refactoring to make the presence (or
lack thereof) of the token not matter.


cassierecher:
CI failure looks unrelated to this code

cassierecher:
Done!

PR #361, opened 2019-10-11
Clone MD parser to create new MD parser and add more tests.

cassierecher:
This is mostly a copy of parser/md with some minor nonfunctional edits to parse.go and parse_test.go, and some added tests to parse_test.go to make it easier to add features later.


cassierecher:
I don't know how long that will take, and depending on how widely used parser/md is, we might not be able to completely get rid of it at all. This duplication also lets us vary the public API, and hopefully remove some of the the things like dependence on the existence of a network/local disk.

PR #362, opened 2019-10-15
fix(event-drowdown): only display events dropdown if events exists

M3lkior:
Using tools for internal usecases, we don't need to use Events features so we only use the default view
If not event views was defined, the event dropdown is displayed
This fix don't display the drop-drop if no views exists


marcacohen (approved):
thanks!

PR #367, opened 2019-11-14
Markdown: Link's inline content

nlepage:
Allow links to have inline elements such as images.

### Example
With the following markdown:
```markdown
[![Alt text](image.jpg)](https://example.com)
```

The link and the image were discarded.

With this PR the HTML output is:
```html
<p>
  <a href="https://example.com">
    <img src="hash.jpg">
  </a>
</p>
```

### DIFF
--- a/claat/parser/md/html.go
+++ b/claat/parser/md/html.go
@@ -46,19 +46,21 @@ func isMeta(hn *html.Node) bool {
 }
 
 func isBold(hn *html.Node) bool {
-	if hn.Type == html.TextNode {
-		hn = hn.Parent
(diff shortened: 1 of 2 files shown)


nlepage:
I'm not so sure about the changes in `isBold`, `isItalic` and `isCode`.

The goal is to keep the same behavior for this type of markdown input:
```markdown
**Some text [a link](href) some more text**
```

It ensures that the link's text is in a `<strong>` element.

jlandure:
Any news on this enhancement @jatorre @KeyboardNerd ? 👍

PR #376, opened 2019-12-20
added Stackblitz and Codesandbox to iframe whitelist

Snickdx:


joshjcarrier:
Since you're modifying this list, would you care to add forms.microsoft.com :) ?

Snickdx:
So... is anyone going to review this?

marcacohen (changes requested):
Hi, can you adjust the offset of the youtube line so it aligns with rest of the block?

PR #385, opened 2020-02-21
Fix Keyboard navigation

nicolasgarnier:


nicolasgarnier:
For the build issue, it is caused by a missing file on the bazel mirror. I filed an issue here: https://github.com/bazelbuild/rules_closure/issues/472

PR #396, opened 2020-03-20
Fix duplicated exported content with memory export

### LINKED ISSUE #391: In Memory Export function results in duplicated codelabs content
To repro: change the fetcher in the Codelab Export to MemoryFetcher, and use that to slurp out the codelab. You'll see that the thing put into the output buffer contains (generated content) * (number of codelab.Step).

KeyboardNerd:
Fixes #391

### DIFF
--- a/claat/cmd/export.go
+++ b/claat/cmd/export.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


claat/cmd/export_test.go
```diff
@@ -41,7 +42,6 @@ func TestExportCodelabMemory(t *testing.T) {
 		{
 			name:     "Multiple Steps",
 			filePath: "testdata/simple-2-steps.md",
-			knownBug: "https://github.com/googlecodelabs/tools/issues/391",
```
shawnbuso: Can we remove the test.knownBug check below now?

claat/cmd/export.go
```diff
@@ -169,6 +169,16 @@ func writeCodelabWriter(w io.Writer, clab *types.Codelab, extraVars map[string]s
 		Steps:    clab.Steps,
 		Extra:    extraVars,
 	}}
+
+	if ctx.Format != "offline" {
+		ext := ctx.Format
```
shawnbuso: Does this end up getting used somewhere? Do we need the intermediate variable?

claat/cmd/export.go
```diff
@@ -169,6 +169,16 @@ func writeCodelabWriter(w io.Writer, clab *types.Codelab, extraVars map[string]s
 		Steps:    clab.Steps,
 		Extra:    extraVars,
 	}}
+
+	if ctx.Format != "offline" {
```
shawnbuso: Does it make more sense to invert this? 

if ctx.Format == "offline"
  for loop
else
  stuff currently in conditional block

KeyboardNerd: sure, either way is fine for me

claat/cmd/export.go
KeyboardNerd: I actually think we should remove the "offline" support for the Memory Export because  offline should require multiple io.Writer instead of one to make it work properly.

claat/cmd/export_test.go
KeyboardNerd: Removed.

PR #413, opened 2020-04-28
Parse <<path>> for markdown imports

KeyboardNerd:
Example is provided in claat/cmd/testdata/import-test.md.

User writes

```
<<path>>
```

in the markdown file to import another markdown fragment file.

Note that if user want to use `<<path>>` literal in the code block, they need to do

\`\`\`
(at least one space here) \<\<path\>\>
\`\`\`

otherwise, it'll be interpreted as import command.

Path can be local file path, relative or absolute, or remote path in Google drive, or gdoc (I think that will work as well).

-----

Side Note on the implementation:
The syntax:

```
<<path>>
```

has a lot of problems with the HTML syntax and character escaping. 

So the solution is to preprocess it and convert this into

```
<!--__unsupported_import_zmcgv2epyv=path-->
```

HTML comment for extremely easy parsing without big change in the HTML parsing process. And if people put this in wrong places e.g. before any step, it'll be hidden because that's a html comment.

__unsupported_import_zmcgv2epyv contains a random string just to yell at people "you should not use this yourself!" and it's subject to be changed in the future if there's a need (I guess not).

### DIFF
--- a/claat/parser/md/parse.go
+++ b/claat/parser/md/parse.go
@@ -66,7 +66,7 @@ const (
 )
 
 var (
-	importsTagRegexp           = regexp.MustCompile("^<<([^<>()]+)>>\\s*$")
+	importsTagRegexp           = regexp.MustCompile("^<<([^<>()]+.md)>>\\s*$")
 	convertedImportsDataPrefix = "__unsupported_import_zmcgv2epyv="
 	convertedImportsPrefix     = []byte("<!--" + convertedImportsDataPrefix)
 	convertedImportsSuffix     = []byte("-->")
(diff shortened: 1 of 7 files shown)


claat/parser/md/parse.go
```diff
@@ -65,6 +65,13 @@ const (
 	headerFAQ   = "frequently asked questions"
 )
 
+var (
+	importsTagRegexp           = regexp.MustCompile("^<<([^<>()]+)>>\\s*$")
```
nicolasgarnier, line 69: I'm thinking we could even restrict it to .md file extensions for the moment. Just to be on the safe side.

KeyboardNerd: Done restricted for now. If user want remote resource e.g. something in google drive, we can relax this later.

Issue #428, opened 2020-07-08
***bold italics*** does not work in md parser


M4FFS:
Steps to reproduce:

1. Create codelab in .md format 
2. Attempt to create text in ***bold italics*** in any of the following ways: 
    `***text***
    ___text___
    **_ text_**
    __* text*__`
3. All will fail and simply render as *bold italics*

M4FFS:
PR created https://github.com/googlecodelabs/tools/pull/429

PR #429, opened 2020-07-08
Add support for bold italics text

M4FFS:
Fix for issue https://github.com/googlecodelabs/tools/issues/428

### DIFF
--- a/claat/parser/md/parse.go
+++ b/claat/parser/md/parse.go
@@ -849,6 +849,7 @@ func link(ds *docState) types.Node {
 func text(ds *docState) types.Node {
 	bold := isBold(ds.cur)
 	italic := isItalic(ds.cur)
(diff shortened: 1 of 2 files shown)


claat/parser/md/parse.go
```diff
@@ -849,6 +849,10 @@ func link(ds *docState) types.Node {
 func text(ds *docState) types.Node {
 	bold := isBold(ds.cur)
 	italic := isItalic(ds.cur)
+	if isBoldAndItalic(ds.cur) {
```
cassierecher, line 852: Add a comment explaining why we need this a separate case from bold and italic individually.

cassierecher, line 852: Add a comment explaining why we need this a separate case from bold and italic individually.

PR #437, opened 2020-07-21
Fix youtube nodes in md renderer

M4FFS:
Previously, when compiling a gdoc codelab to markdown, embedded youtube videos would be ignored

Now they are properly rendered using the chosen youtube syntax defined in https://github.com/googlecodelabs/tools/pull/400

### DIFF
--- a/claat/render/md.go
+++ b/claat/render/md.go
@@ -114,8 +114,8 @@ func (mw *mdWriter) write(nodes ...types.Node) error {
 		//	mw.survey(n)
 		case *types.HeaderNode:
 			mw.header(n)
-			//case *types.YouTubeNode:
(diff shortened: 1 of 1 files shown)


claat/render/md.go
```diff
@@ -262,3 +262,10 @@ func (mw *mdWriter) header(n *types.HeaderNode) {
 		mw.writeBytes(newLine)
 	}
 }
+
+func (mw *mdWriter) youtube(n *types.YouTubeNode) {
+	mw.newBlock()
+	mw.writeString(`<video id="`)
```
cassierecher, line 268: Could this be one writeString call using Sprintf instead?

M4FFS: Good call, done

PR #379, opened 2020-01-30
Enable markdown buttons

kroikie:
This PR enables the `isButton` check in the md html parser. It checks if the node in question is a Button atom.

PR #439, opened 2020-07-28
Update md parser README

M4FFS:


claat/parser/md/README.md
```diff
@@ -108,9 +124,46 @@ This will appear in a negative info box.
 
 Codelabs sometimes contain links to SDKs or sample code. The codelab renderer
 will apply special button-esque styling to any link that begins with the word
-"Download".
```
shawnbuso, line 111: This is a little confusing to me - are we saying that either of these is an option? If so maybe we should make them 2 separate code blocks, like:

you can do this:
```
[Download thing]()
```
or this:
```
<button>
  [Download thing]()
</button>
```

And now that I type that, I'm a little more confused - doesn't the second example contain the first example? What's the difference?

M4FFS: Good catch! I forgot to update the text which was wrong, currently the only way to render download buttons is [with a <button>](https://github.com/googlecodelabs/tools/pull/379#issuecomment-613709303), I will update the text to reflect that, or, alternatively, try make it work with regular links?

claat/parser/md/README.md
```diff
@@ -108,9 +124,46 @@ This will appear in a negative info box.
...
+#### "Frequently Asked Questions" Section
+
+If you add a FAQ section to your codelab in the following format, then respective
+link icons will be automatically added to replace the bullet point symbol. (Note
+that the "Frequently Asked Questions" string must be exact)
+
+```
+### Frequently Asked Questions
+* [stack overflow link](example.com/)
+* [developers.android.com link ](example.com/)
```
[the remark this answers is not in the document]
MoonWolf125: Just below these is the Support site and I've never seen it's icon work in the FAQ section. Is it working for others?

Issue #422, opened 2020-06-30
claat/fetch_test.go failing on windows


M4FFS:
Currently, fetch_test.go FAILs on windows in a freshly cloned repo despite it passing in a linux environment 

**Steps to reproduce:**
1. Clone repo on a windows machine 
_(powershell:)_
2. `cd` into `tools/`
3.  `git test ./...` 
4. Test case fails:

Produced in Windows 10 Home 64-bit

```
--- FAIL: TestRestrictPathToParent (0.00s)
    --- FAIL: TestRestrictPathToParent/asset:_/tmp/imgabs.png,_parent:_foo/ (0.00s)
        fetch_test.go:80: restrictPathToParent() error = <nil>, wantErr true
FAIL
FAIL    github.com/googlecodelabs/tools/claat/fetch     0.185s
```

M4FFS:
Any ideas on why the path strings look like this?

![image](https://user-images.githubusercontent.com/40577281/88886708-fba03600-d232-11ea-915f-daec84b7d813.png)

PR #441, opened 2020-07-30
Modify file sources in Makefile to work with windows 

M4FFS:
Also removed hardcoded `render/tmpldata.go` from SRS variable in Makefile as it was being correctly found by the `find` command


M4FFS:
This does not fix https://github.com/googlecodelabs/tools/issues/422 but it makes some progress towards it

PR #446, opened 2020-08-11
Add github to allowlist of iframe domains

julioz:
Allowing for codelabs to render GitHub pages might be useful to refer to other code or documentation pages (READMEs, for instance).


julioz:
Hi @cassierecher sorry to direct ping, but this feature would be super useful to us at SoundCloud; as we would like to prevent copy-pasting sample source code into our codelabs and instead point to our already existing sources in production codebases. Do you reckon merging this PR is enough to allow for that?

marcacohen (approved):
thanks!

PR #448, opened 2020-08-13
Fixes incorrect markdown button rendering.

shawnbuso:


claat/render/md.go
```diff
@@ -172,17 +172,21 @@ func (mw *mdWriter) image(n *types.ImageNode) {
...
-		mw.writeString("[")
-	}
-	for _, cn := range n.Content.Nodes {
-		if t, ok := cn.(*types.TextNode); ok {
-			mw.writeString(t.Value)
+		// Look-ahead for button syntax.
+		if _, ok := n.Content.Nodes[0].(*types.ButtonNode); ok {
+			mw.writeString("<button>")
 		}
+		mw.writeString("[")
```
cassierecher, line 179: Is this guaranteed to always be closed for reasonable input?

shawnbuso: I think it should be - I close it on line 183 below under the same condition (`n.URL != ""`), so the only thing that should stop it from being closed is if we for some reason abort writing the markdown in the nested `write` calls. But that would break a whole lot more than just this.

PR #449, opened 2020-08-14
Write infoboxes in vanilla markdown instead of using HTML aside.

shawnbuso:


claat/render/md.go
```diff
@@ -243,24 +243,22 @@ func (mw *mdWriter) itemsList(n *types.ItemsListNode) {
...
-	// at the beginning. This is because the gdoc parser parses info boxes as a list node
-	// followed by a text node, and the list node is an md block unto itself, resulting in two
-	// newlines being added to the output before the text.
+	// InfoBoxes are comprised of a ListNode with the contents of the InfoBox. 
+	// Writing the ListNode directly results in extra newlines in the md output
+	// which breaks the formatting. So instead, write the ListNode's children 
+	// directly and don't write the ListNode itself.
+	ln, ok := n.Content.Nodes[0].(*types.ListNode)
+	if !ok {
+		return
```
shawnbuso, line 252: I feel like I need to do something better here, but we don't have a pattern for write methods raising errors. Thoughts?

claat/render/md.go
```diff
@@ -243,24 +243,22 @@ func (mw *mdWriter) itemsList(n *types.ItemsListNode) {
...
-	//
-	// The main issue is that when you do write(n.Content.Nodes...) it always adds two newlines
-	// at the beginning. This is because the gdoc parser parses info boxes as a list node
-	// followed by a text node, and the list node is an md block unto itself, resulting in two
-	// newlines being added to the output before the text.
+	// InfoBoxes are comprised of a ListNode with the contents of the InfoBox. 
+	// Writing the ListNode directly results in extra newlines in the md output
+	// which breaks the formatting. So instead, write the ListNode's children 
+	// directly and don't write the ListNode itself.
+	ln, ok := n.Content.Nodes[0].(*types.ListNode)
```
shawnbuso, line 250: I actually have two options here:

1) write the ListNode's children directly, and don't write the ListNode itself
2) modify the ListNode so it's no longer a block, and then write the ListNode itself

What's implemented here is (1). Is that better than (2)?

shawnbuso: I've thought about it a bit and I think (2) is more work than it's worth, and I can't find a good reason that it's better than (1), so I'm just gonna stick with (1).

claat/render/md.go
cassierecher: You could add something like docState (see the MD parser) that would report the errors at the end, but given how large an undertaking that would be, I don't think it's necessarily required here. Please do file an issue about it though, we should have better error handling overall.

shawnbuso: Done - https://github.com/googlecodelabs/tools/issues/452

Issue #443, opened 2020-08-04
Latest version ignores steps


haz:
I've updated to the latest version (from the master branch) to try and get the button functionality working (using the `<button>` tags), but I'm finding that the sections are now ignored, and just one big step is created. Following [this tutorial](https://medium.com/@zarinlo/publish-technical-tutorials-in-google-codelab-format-b07ef76972cd), the example markdown file has the same behaviour (so I'm pretty sure it's not something I've messed up with my own example.

Using the markdown version, and compiling with `claat export -o build/ install.md`. Any pointers on what I might be doing wrong would be greatly appreciated.

Issue #315, opened 2019-07-19
blackfriday.v2 parser breaks markdown code snippet


Mrc0113:
The changes in this commit cause the markdown parser to no longer properly handle code snippets. 

https://github.com/googlecodelabs/tools/commit/b3953a6ef12a5a4889cc064202af5b4576b77ca5#diff-f622cf9001c4cbee020860ffc936bb0c

The "correct.png" is what was created before this change; now that same markdown file shows what is is "incorrect.png". You'll notice the code snippet doesn't highlight and it also messes up formatting after the code snippet. (In this case a Header 3, but in other cases I have seen it mess up bullets too) 

### Verify PubSub+ Service in Pivotal Cloud Foundry (PCF)
If you are using PCF, your administrator will have created an org and space for your workshop demo in which you can deploy and run your microservices.

**CORRECT.PNG**
![correct](https://user-images.githubusercontent.com/1815312/61540645-0911c600-aa0c-11e9-9afd-3cfec1fa3a85.png)

**INCORRECT.PNG**
![incorrect](https://user-images.githubusercontent.com/1815312/61540646-0911c600-aa0c-11e9-84df-a564f6ff0152.png)

Mrc0113:
FYI - thread opened in authors group here: https://groups.google.com/forum/#!topic/codelab-authors/CgWVZd4bI9Q

marcacohen:
@cassierecher Cassie - looks like your recent PR broke this. Can you take a look?

M4FFS:
I am investigating the cause of this, however a workaround for now that works is adding a blank line before the start of the code block: 

e.g. the following seems to work 
```
* bullet point 1

```
code block
with 
multiple lines
```
* bullet point 2

PR #458, opened 2020-09-11
Change md parser to goldmark

M4FFS:
### The case for goldmark

Multiple issues relating to the use of whitespace have arisen, including [code blocks not rendering properly](https://github.com/googlecodelabs/tools/issues/315), lists not rendering properly when interspersed with links and code blocks, tables sometimes refusing to render and [steps being completely ignored](https://github.com/googlecodelabs/tools/issues/443) etc. It was determined that while it is possible to create workaround fixes for these kinds of bugs, the core issue was problems from the [blackfriday parser](https://github.com/russross/blackfriday). A number of different alternative parsers were tried, and the most success was had with [goldmark](https://github.com/yuin/goldmark)

Unfortunately due to an known regression bug, replacing blackfriday with goldmark on the latest version of master does not produce the intended results, and I assume this is due to recent changes to the parser to attempt to fix bugs caused by the blackfriday parser. I believe that switching to goldmark would be better for the long term success of claat, as it seems to be more reliable and also sports a number of useful parsing features which can be explored on it's github page. 

In this diff, two extensions are loaded into the goldmark parser; `Table` and `Typographer`. `Table` is self explanatory, `Typographer` is similar to blackfriday's `smartypants` feature. 

[Blackfriday parsing test result](https://i.imgur.com/OMGIfef.png) (image link)

[Goldmark parsing test result](https://i.imgur.com/316XK9h.png) (image link)

### DIFF
--- a/claat/parser/md/parse.go
+++ b/claat/parser/md/parse.go
@@ -30,13 +30,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/yuin/goldmark"
+	"github.com/yuin/goldmark/extension"
+
 	"golang.org/x/net/html"
 	"golang.org/x/net/html/atom"
 
 	"github.com/googlecodelabs/tools/claat/parser"
 	"github.com/googlecodelabs/tools/claat/types"
 	"github.com/googlecodelabs/tools/claat/util"
-	"gopkg.in/russross/blackfriday.v2"
 )
 
 // Metadata constants for the YAML header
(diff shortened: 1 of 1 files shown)


M4FFS:
[Blackfriday parsing test result](https://i.imgur.com/OMGIfef.png) (image link)

[Goldmark parsing test result](https://i.imgur.com/316XK9h.png) (image link)

shawnbuso:
"Unfortunately due to an known regression bug, replacing blackfriday with goldmark on the latest version of master does not produce the intended results, and I assume this is due to recent changes to the parser to attempt to fix bugs caused by the blackfriday parser."

Can we do some more investigation here to make sure we can easily fix these?

cassierecher:
I like this but we need to retain the ability to use blackfriday in case there are already-existing codelabs that don't work with goldmark. I'll experiment with this and open a separate PR.

PR #468, opened 2020-09-17
Fix incorrect import branch selector.

cassierecher:
Also remove some parentheses that don't need to be there.

### DIFF
--- a/claat/parser/md/parse.go
+++ b/claat/parser/md/parse.go
@@ -36,7 +36,7 @@ import (
 	"github.com/googlecodelabs/tools/claat/parser"
 	"github.com/googlecodelabs/tools/claat/types"
 	"github.com/googlecodelabs/tools/claat/util"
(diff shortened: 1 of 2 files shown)


claat/parser/md/parse.go
```diff
@@ -36,7 +36,7 @@ import (
 	"github.com/googlecodelabs/tools/claat/parser"
 	"github.com/googlecodelabs/tools/claat/types"
 	"github.com/googlecodelabs/tools/claat/util"
-	"github.com/russross/blackfriday/v2"
+	"github.com/russross/blackfriday.v2"
```
nicolasgarnier, line 39: I think you also need to change go.mod

nicolasgarnier (changes requested):

claat/parser/md/parse.go
nicolasgarnier: That may be why travis is failing...

PR #475, opened 2020-09-21
Fix alt text, titles in images

DavidScales:
Previously, images with titles or containing newlines in alt text
would not render correctly

This commit
* removes newlines in alt text, which break markdown syntax (for
consecutive newlines) and don't parse correctly (for single newlines)
* removes duplicate double quotes applied to title attributes, which
break html syntax
* removes author-added double quotes in titles, which also break html
syntax, and replace with `&quot;`

### DIFF
--- a/claat/parser/gdoc/parse.go
+++ b/claat/parser/gdoc/parse.go
@@ -622,7 +622,7 @@ func code(ds *docState, term bool) types.Node {
 	} else if ds.cur.Parent.FirstChild == ds.cur && ds.cur.Parent.DataAtom != atom.Span {
 		v = "\n" + v
 	}
-	var lang string;
+	var lang string
 	n := types.NewCodeNode(v, term, lang)
 	n.MutateBlock(td)
 	return n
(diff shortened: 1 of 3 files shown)


claat/render/md.go
```diff
@@ -174,7 +174,7 @@ func (mw *mdWriter) image(n *types.ImageNode) {
 	}
 
 	if n.Title != "" {
-		mw.writeString(fmt.Sprintf("title=\"%q\" ", n.Title))
+		mw.writeString(fmt.Sprintf("title=\"%s\" ", n.Title))
```
cassierecher, line 177: Does the fix still work if it's `title=%q` instead?

claat/parser/gdoc/parse.go
```diff
@@ -701,9 +705,10 @@ func image(ds *docState) types.Node {
 	if errorAlt != "" {
 		n.Alt = errorAlt
 	} else {
-		n.Alt = nodeAttr(ds.cur, "alt")
+		n.Alt = alt
 	}
-	n.Title = nodeAttr(ds.cur, "title")
+	// Author-added double quotes in attributes break html syntax
+	n.Title = strings.Replace(nodeAttr(ds.cur, "title"), `"`, "&qout;", -1)
```
cassierecher, line 711: Should this be quot?

cassierecher: (another instance of this on L675)

cassierecher (changes requested):

claat/render/md.go
DavidScales: no, stripping quotes in the parser only corrects for author-added quotes, but this line in the renderer is incorrect by itself because the `%q` symbol in Go adds quotes, so `title=\"%q\" ` will always render `title=""some title""`

DragonRider0o0:
Comment seems outdated:
_Removes author-added double quotes in titles, which also break html
syntax, and replace with single quotes_

Should say replace with `&qout;`.

claat/render/md.go
DavidScales: yes and that's cleaner, I'll switch over to that

claat/parser/gdoc/parse.go
DavidScales: oh this is embarrassing... I'll fix that

cassierecher (approved):

PR #494, opened 2020-10-05
Fix broken build.

shawnbuso:
We are on too old a version of closure to use goog.asserts.dom. We need to update to at least 20191111, but that required a closure compiler update, which I think required a bazel update, which made all of our configs invalid, etc., etc., etc. The real fix here is to update to the latest version of bazel and the closure compilers, but that will be a project in and of itself, so this is a quick fix to unblock us.


shawnbuso:
Superseded by #498

PR #506, opened 2020-10-09
Gfm tables

narphorium:


claat/render/md.go
```diff
@@ -323,21 +324,46 @@ func (mw *mdWriter) youtube(n *types.YouTubeNode) {
...
+	// Calculate table inner content length.
+	contentLength := 0
+	for _, row := range n.Rows {
+		for _, cell := range row {
+			var nw bytes.Buffer
+			WriteMD(&nw, mw.env, mw.format, cell.Content.Nodes...)
+			contentLength += len(bytes.Trim(nw.Bytes(), " \t"))
+		}
+	}
+	// If table content is empty, don't output the table.
```
DragonRider0o0, line 336: Please use `GridNode.Empty()`.
See https://github.com/googlecodelabs/tools/blob/master/claat/types/node.go#L227-L237

claat/render/md.go
```diff
@@ -323,21 +324,46 @@ func (mw *mdWriter) youtube(n *types.YouTubeNode) {
...
+
+			// Check cell content for newlines and replace with inline HTML if newlines are present.
+			var nw bytes.Buffer
+			WriteMD(&nw, mw.env, mw.format, cell.Content.Nodes...)
+			if bytes.ContainsRune(nw.Bytes(), '\n') {
+				for _, cn := range cell.Content.Nodes {
+					cn.MutateBlock(false) // don't treat content as a new block
+					var nw2 bytes.Buffer
+					WriteHTML(&nw2, mw.env, mw.format, cn)
+					mw.writeBytes(bytes.ReplaceAll(nw2.Bytes(), []byte("\n"), []byte("")))
```
DragonRider0o0, line 357: Consider replacing new lines with `<br>` or ` ` rather than just removing them. <br> is probably the best path forward.

DragonRider0o0 (commented):
See my comments

- https://github.com/googlecodelabs/tools/pull/506/files#r502498143

- https://github.com/googlecodelabs/tools/pull/506/files#r502500878

claat/render/md.go
shawnbuso: I think in this case it's newlines in the markdown source that we're replacing, so we probably don't want <br />.

e.g. we are changing this:
<ul>
  <li>thing</li>
</ul>

to this:
<ul><li>thing</li></ul>

If the goal is to preserve newlines in text, we should probably add something to the text render method that replaces them with <br />, but leave this as-is.

DragonRider0o0: Shawn is right, replacing `\n` with `<br>` would be better placed in text rendering logic.

shawnbuso: I think we can do that in another PR - but Shawn should address the other comment above before merging.

shawnbuso (approved):

PR #530, opened 2020-12-10
Fix issues rendering unmergeable adjacent bold and italic nodes.

shawnbuso:


claat/parser/trim.go
```diff
@@ -174,6 +174,21 @@ func splitSpaceRight(s string) (v string, sp string) {
...
 
+func requiresSpacer(a, b types.Node) bool {
+	t1, ok1 := a.(*types.TextNode)
+	t2, ok2 := b.(*types.TextNode)
+
+	if !(ok1 && ok2) {
+		return false
+	}
+
+	if (t1.Bold && t1.Bold == t2.Bold) || (t1.Italic && t1.Italic == t2.Italic) {
```
cassierecher, line 185: Return this expression instead of using a conditional

shawnbuso: Ugh duh - fixed.

claat/parser/md/html.go
```diff
@@ -56,6 +65,15 @@ func isBold(hn *html.Node) bool {
...
+		// Look up as many as 2 levels, to handle the case of e.g. <em><bold><code>
+		for i:= 0; i < 2; i++ {
+			hn = hn.Parent
+			if hn.DataAtom == atom.Em || hn.DataAtom == atom.I {
+				return true
+			}
+		}
+		return false
 	}
 	return hn.DataAtom == atom.Em ||
```
cassierecher, line 78: Could this be moved into the `if` block on L66?

claat/parser/trim.go
```diff
@@ -174,6 +174,17 @@ func splitSpaceRight(s string) (v string, sp string) {
...
 
+func requiresSpacer(a, b types.Node) bool {
+	t1, ok1 := a.(*types.TextNode)
+	t2, ok2 := b.(*types.TextNode)
+
+	if !(ok1 && ok2) {
+		return false
+	}
+
+	return (t1.Bold && t1.Bold == t2.Bold) || (t1.Italic && t1.Italic == t2.Italic)
```
cassierecher, line 185: If I read right, this is asking "are `t1` and `t2` either both bold or both italic?" I think it would be more readable to test those directly instead of using `==`.

cassierecher (changes requested):

claat/parser/md/html.go
shawnbuso: I could replicate it there but I can't move it there - I need it here to handle non-text nodes and non-code nodes.

claat/parser/trim.go
shawnbuso: Yeah this is way better - fixed.

claat/parser/md/html.go
cassierecher: Ah, I misread `else if` as `else`.

cassierecher (approved):

PR #554, opened 2021-04-06
Add Go Modules support for v2.2.4

atavakoliyext:
The differences between this and the configuration provided in
claat/v2.2.3:

- This was built on tag v2.2.4 (via go mod init and go mod tidy)
- This defines the name with the required major-version suffix (/v2)


atavakoliyext:
The main motivation for this was to allow the claat module to be usable as a dependency in other modules, which is currently not possible due to the module name not having the correct major-version suffix, producing errors of the form:

```
go: github.com/googlecodelabs/tools/claat/v2@v2.2.3: claat/go.mod has non-.../v2 module path "github.com/googlecodelabs/tools/claat" (and .../v2/go.mod does not exist) at revision claat/v2.2.3
```

To avoid the issues that motivated reverting `claat/v2.2.3`, would it be possible to have this commit live on a different branch created at tag v2.2.4? I don't believe there's a way to create a new branch in a PR, but I can edit my PR if this approach is deemed ok.

Hopefully at some point module support can exist on the master branch.

Rada8732 (approved):

Rada8732:
Why are checks failing

Rada8732:
rogchap.com/v8go

cassierecher:
I was able to create a module for the project on the master branch, though it does not make reference to the version number. I am not very familiar with the Go modules system; is this sufficient for your use case?

nievesmiguel (approved):

PR #577, opened 2021-06-09
Create Go module.

cassierecher:


shawnbuso (approved):

cassierecher:
Some context notes:

- This module purposefully is not aware that the code is on v2. I am not certain, but expect this to not be an issue as v1 is not available as a module. If we decide we need v3 in the future, we can start versioning then.
- I will seek alternative solutions to the problems that caused us to revert the previous attempt at modularization.

PR #594, opened 2021-06-30
Add tests for InfoboxNode.

cassierecher:


claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+// AllowUnexported option for cmp to make sure we can diff properly.
+var cmpOpt = cmp.AllowUnexported(InfoboxNode{}, node{}, ListNode{}, TextNode{})
+
+func TestNewInfoboxNode(t *testing.T) {
+	tests := []struct {
+		name      string
+		inKind    InfoboxKind
```
DragonRider0o0, line 15: Consider dropping the `in` prefix; `kind` and `content` are sufficient.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+			// TODO: Should we set a default value?
+			name:      "NoKind",
+			inContent: []Node{NewTextNode("orange")},
+			out: &InfoboxNode{
+				node:    node{typ: NodeInfobox},
+				Content: NewListNode(NewTextNode("orange")),
+			},
+		},
+		{
+			// TODO: Should a list of exactly one list be flattened?
```
DragonRider0o0, line 89: I don't think we should flatten single item lists, I think that's best left as an author decision rather than a platform decision.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+			inKind:    InfoboxNegative,
+			inContent: []Node{NewTextNode("orange"), NewTextNode("strawberry"), NewTextNode("pineapple")},
+			out: &InfoboxNode{
+				node:    node{typ: NodeInfobox},
+				Kind:    InfoboxNegative,
+				Content: NewListNode(NewTextNode("orange"), NewTextNode("strawberry"), NewTextNode("pineapple")),
+			},
+		},
+		{
+			// TODO: Should we set a default value?
```
DragonRider0o0, line 80: We should probably set the default value to positive.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+		inContent []Node
+		out       *InfoboxNode
+	}{
+		{
+			name:   "PositiveEmpty",
+			inKind: InfoboxPositive,
+			out: &InfoboxNode{
+				node: node{typ: NodeInfobox},
+				Kind: InfoboxPositive,
+				// TODO: Do we really want this to not be nil?
```
DragonRider0o0, line 25: I think it's acceptable as is because we're calling the NewListNode's empty method when we evaluate empty.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+				Content: NewListNode(NewTextNode("orange"), NewTextNode("strawberry"), NewTextNode("pineapple")),
+			},
+		},
+		{
+			name:   "NegativeEmpty",
+			inKind: InfoboxNegative,
+			out: &InfoboxNode{
+				node: node{typ: NodeInfobox},
+				Kind: InfoboxNegative,
+				// TODO: Do we really want this to not be nil?
```
DragonRider0o0, line 55: I think it's acceptable as is because we're calling the NewListNode's empty method when we evaluate empty.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+			inContent: []Node{NewTextNode("orange"), NewTextNode("strawberry"), NewTextNode("pineapple")},
+			out: &InfoboxNode{
+				node:    node{typ: NodeInfobox},
+				Kind:    InfoboxNegative,
+				Content: NewListNode(NewTextNode("orange"), NewTextNode("strawberry"), NewTextNode("pineapple")),
+			},
+		},
+		{
+			// TODO: Should we set a default value?
+			name:      "NoKind",
```
DragonRider0o0, line 81: We should probably add a test for an unsupported kind as well.

claat/nodes/infobox_test.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			out := NewInfoboxNode(tc.inKind, tc.inContent...)
+			if diff := cmp.Diff(tc.out, out, cmpOpt); diff != "" {
+				t.Errorf("NewInfoboxNode(%q, %v) got diff (-want +got): %s", tc.inKind, tc.inContent, diff)
+				return
+			}
+		})
+	}
+}
```
DragonRider0o0, line 109: Add a `TestInfoboxNodeEmpty` test to validate `Empty`.

claat/nodes/infobox_test.go
cassierecher: It would just be flattening in the case of a list containing exactly one list. I'm not sure it matters in terms of output; this may be an optimization we can make before rendering.

claat/nodes/infobox_test.go
cassierecher: Probably, but want to leave this for a future PR and have this one just contain tests for the code as it is today.

claat/nodes/infobox_test.go
cassierecher: I think this eventually ties in to a look at our node system. I'm not sure we have a universal standard over whether an empty node is semantically synonymous to the lack of a node or not.

claat/nodes/infobox_test.go
cassierecher: We could just as easily check if Content is nil, or perhaps in addition to. Would like to defer this to a separate PR

claat/nodes/infobox_test.go
cassierecher: I think this makes it more readable, it provides more cues to what the test's inputs and outputs are.

claat/nodes/infobox_test.go
DragonRider0o0: SGTM.

claat/nodes/infobox_test.go
DragonRider0o0: Fair point.

claat/nodes/infobox_test.go
DragonRider0o0: SGTM

claat/nodes/infobox_test.go
DragonRider0o0: Right, but the first list represents the content of the infobox while the second list represents a list within the content of the infobox.

cassierecher: Ah, I understand now. Will remove this

claat/nodes/infobox_test.go
cassierecher: Added

claat/nodes/infobox_test.go
cassierecher: Added

claat/nodes/infobox_test.go
cassierecher: Parallel discussion on https://github.com/googlecodelabs/tools/pull/596#discussion_r666499324

shawnbuso (approved):

PR #595, opened 2021-06-30
Add tests for YoutubeNode.

cassierecher:


claat/nodes/youtube_test.go
```diff
@@ -0,0 +1,71 @@
...
+			in:   "Mlk888FiI8A",
+			out: &YouTubeNode{
+				node:    node{typ: NodeYouTube},
+				VideoID: "Mlk888FiI8A",
+			},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			out := NewYouTubeNode(tc.in)
```
DragonRider0o0, line 32: Consider renaming `in` to `videoID`

claat/nodes/youtube_test.go
```diff
@@ -0,0 +1,71 @@
...
+				return
+			}
+		})
+	}
+}
+
+func TestYouTubeNodeEmpty(t *testing.T) {
+	tests := []struct {
+		name string
+		in   *YouTubeNode
```
DragonRider0o0, line 44: Use the same pattern for both tests, either pass the video ID string in both or pass the YouTubeNode in both.

cassierecher: Discussion on similar comment https://github.com/googlecodelabs/tools/pull/596/files/591b1ac68607383d635229cd063ed8561fc4b93d

claat/nodes/youtube_test.go
cassierecher: Discussion on similar comment https://github.com/googlecodelabs/tools/pull/596/files/591b1ac68607383d635229cd063ed8561fc4b93d

shawnbuso (approved):

PR #596, opened 2021-06-30
Add tests for IframeNode.

cassierecher:


claat/nodes/iframe_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestNewIframeNode(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
```
DragonRider0o0, line 12: Consider renaming `in` to `url`.

claat/nodes/iframe_test.go
```diff
@@ -0,0 +1,71 @@
...
+				return
+			}
+		})
+	}
+}
+
+func TestIframeNodeEmpty(t *testing.T) {
+	tests := []struct {
+		name string
+		in   *IframeNode
```
DragonRider0o0, line 44: Use the same pattern for both tests, either pass the url string in both or pass the IFrameNode in both.

claat/nodes/iframe_test.go
cassierecher: I find this makes it clearer at a glance what the inputs and outputs are. What about `inURL`?

claat/nodes/iframe_test.go
cassierecher: In the first test, the New function is what's being tested. In the second one, I want to avoid using New to isolate failures if New stops working.

claat/nodes/iframe_test.go
DragonRider0o0: I would prefer `inURL` over `in` and `url` over either. I trust your judgement.

claat/nodes/iframe_test.go
DragonRider0o0: I'm not convinced that's desirable. `NewIframeNode` is the client interface we want to support, I think it makes sense to test with that rather than `IframeNode`'s directly.

claat/nodes/iframe_test.go
cassierecher: inURL works as a compromise I think. I'll propagate that format over the three open PRs as well.

claat/nodes/iframe_test.go
cassierecher: Good point. Done

shawnbuso (approved):

PR #597, opened 2021-06-30
Add tests for SurveyNode.

cassierecher:


claat/nodes/survey_test.go
```diff
@@ -0,0 +1,210 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestNewSurveyNode(t *testing.T) {
+	tests := []struct {
+		name     string
+		inID     string
```
DragonRider0o0, line 12: Consider dropping the `in` prefix; `id` and `groups` are sufficient.

claat/nodes/survey_test.go
```diff
@@ -0,0 +1,210 @@
...
+		out      *SurveyNode
+	}{
+		{
+			name: "Empty",
+			out: &SurveyNode{
+				node: node{typ: NodeSurvey},
+			},
+		},
+		{
+			// TODO: should the absence of an ID be an error?
```
DragonRider0o0, line 23: Good question, I think it should either evaluate as empty or as a valid depending on the desired behavior for SurveyNode.

claat/nodes/survey_test.go
```diff
@@ -0,0 +1,210 @@
...
+				return
+			}
+		})
+	}
+}
+
+func TestSurveyNodeEmpty(t *testing.T) {
+	tests := []struct {
+		name string
+		in   *SurveyNode
```
DragonRider0o0, line 128: Use the same pattern for both tests, either pass the groups and ID in both or pass the SurveyNode in both.

cassierecher: Discussion on similar comment https://github.com/googlecodelabs/tools/pull/596/files/591b1ac68607383d635229cd063ed8561fc4b93d

claat/nodes/survey_test.go
cassierecher: Discussion on similar comment https://github.com/googlecodelabs/tools/pull/594/files/8acfb33186a4e580c372a34d41087c721343dd3d#r666514627

claat/nodes/survey_test.go
cassierecher: Yeah, we need to figure out what the desired behavior should be. I would like to defer this to a separate PR

DragonRider0o0: SGTM

shawnbuso (approved):

PR #600, opened 2021-07-01
Strip spaces from keywords and normalize keyword parsing

DragonRider0o0:


claat/util/util.go
```diff
@@ -30,3 +30,28 @@ func Unique(a []string) []string {
...
+	}
+	s = stripSpaces(s)
+	strs := strings.Split(s, ",")
+	for k, v := range strs {
+		strs[k] = strings.ToLower(v)
+	}
+	return Unique(strs)
+}
+
+func stripSpaces(s string) string {
```
cassierecher, line 48: Could we make use of http://go/godoc/strings/#TrimSpace instead?

claat/util/util_test.go
```diff
@@ -37,3 +37,55 @@ func TestUnique(t *testing.T) {
 		})
 	}
 }
+
+func TestNormalizedSplit(t *testing.T) {
```
cassierecher, line 41: Add a test case with some Unicode-defined spaces that are not U+0020

cassierecher, line 41: Add a test case with multiple spaces consecutively

cassierecher (dismissed):

claat/util/util.go
DragonRider0o0: No, we cannot.  We need to remove all of the spaces, not just leading and trailing.

claat/util/util_test.go
DragonRider0o0: Done

DragonRider0o0: Done

shawnbuso (approved):

PR #654, opened 2021-10-05
Add DartPad.dev to the iframe allow list

domesticmouse:
I'd posit that [DartPad](https://dartpad.dev) is on a similar level to codepen.io, glitch.com and repl.it, but maintained by googlers. Our community members are making codelabs with CLAAT targeted at teaching users with DartPad, and it'd be great to enable embedding DartPad directly.

Thanks!


domesticmouse:
PTAL

shawnbuso (approved):

PR #688, opened 2021-11-24
Remove non-existing selector ::content

devbridie:
https://www.w3.org/TR/css-pseudo-4/ claims that this selector doesn't exist. This malformed rule causes the entire block to be skipped, which breaks the checklist list item markers in e.g. "What you'll learn" in https://codelabs.developers.google.com/codelabs/arcore-cloud-anchors#0.


devbridie:
Oh, I think this is an old polymer convention concerning scoped styles: https://polymer-library.polymer-project.org/1.0/docs/devguide/styling. I think this was an oversight in migration.

shawnbuso (approved):

PR #693, opened 2021-12-25
install go module with `go install` instead of `go get`

j143:
for more: https://go.dev/doc/go-get-install-deprecation

### DIFF
--- a/claat/README.md
+++ b/claat/README.md
@@ -14,7 +14,7 @@ The binaries, as well as their checksums are available at the
 
 Alternatively, if you have [Go installed](https://golang.org/doc/install):
 
-    go get github.com/googlecodelabs/tools/claat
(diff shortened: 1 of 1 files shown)


Arry-boy:
`import google  docs  into the code line 42  ` @j143 the user in the app will only get access  to the google personal information  like sharing the data with the merging 
if i am using that app i can see the  google user data

j143:
> `import google docs into the code line 42` @j143 the user in the app will only get access to the google personal information like sharing the data with the merging
if i am using that app i can see the google user data

Hi @Arry-boy -  Sorry, I didn't quite follow. `go get` in the as shown in the docs does not work, hence updated.

cassierecher (approved):

cassierecher:
Thank you!

PR #713, opened 2022-03-09
Fix byte index out of range panic for likely error responses.

shawnbuso:


DragonRider0o0 (approved):
Consider printing the bytes, they might contain a brief error message.

claat/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
```
Groangfr, line 5: Tấc cả ngôn ngữ

PR #798, opened 2022-11-06
fix oob oauth to use localhost and update elements bucket

mco-gh:


claat/fetch/drive/auth/auth.go
```diff
@@ -215,9 +235,9 @@ func (c *cachedTokenSource) Token() (*oauth2.Token, error) {
 // authorize performs user authorization flow, asking for permissions grant.
 func authorize(conf *oauth2.Config) (*oauth2.Token, error) {
 	aurl := conf.AuthCodeURL("unused", oauth2.AccessTypeOffline)
```
cassierecher, line 237: nit: `aURL`

cassierecher (approved):

claat/fetch/drive/auth/auth.go
mco-gh: fixed,PTAL

mco-gh: ah, ignore the ptal, I see you pre-approved, thank you!

PR #799, opened 2022-11-09
Use exact match with IframeAllowlist

DragonRider0o0:


mco-gh (approved):
It's hard to know for certain whether this will break anything but in the spirit of improving security, I can see trying this and adjusting it if we hear any complaints.

PR #804, opened 2022-11-23
Add codesandbox.io To The iframe Allow List

harrismcc:
Tiny PR that adds codesandbox.io to the iframe allow list


harrismcc:
Bumping this

PR #853, opened 2023-09-20
Add GA4 id parser for Google document

khammami:
GA4 was only added for md files


khammami:
Hey @mco-gh, could you please check this PR? (it seems authors group is down because of spams)

GA4 has been added to md parser but not to gdocs parser. (https://github.com/googlecodelabs/tools/commit/4c98edf2e6e50c0687a4cf56ea204232746a92be)

For now I'm using a little script in my gh action to add the missing ga4id.

one more thing, claat binary  doesn't contain the latest commits that support ga v4. any plan for another release for GitHub? (installing it though go will contain the latest changes)

Issue #862, opened 2023-10-28
Continously publish `latest` release binaries


vorburger:
It would be neat if https://github.com/googlecodelabs/tools/releases on top contained a continously published `latest` release.

Hello Project Maintainers: I volunteer to work on this, if you would like to assign this issue to me?

PR #863, opened 2023-10-28
Add Makefile test coverage to GitHub Action Workflow CI

vorburger:
part of #862

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,3 +30,7 @@ jobs:
     - name: Test
       working-directory: claat
       run: go test -v ./...
+
+    - name: Make
+      working-directory: claat
+      run: make all release


vorburger:
Oh, the GitHub Action Workflow CI as-is only runs on `main` and therefore can't test if this actually works...

... let me first raise another PR to propose running it on all PRs, and then we'll see if this actually works.

.github/workflows/go.yml
```diff
@@ -30,3 +30,7 @@ jobs:
     - name: Test
       working-directory: claat
       run: go test -v ./...
+
+    - name: Make
+      working-directory: claat
+      run: make all release
```
vorburger, line 36: Or should this be using `-C` and/or `-j`, as it used to [here](https://github.com/googlecodelabs/tools/blob/9f2b585333fc582b0a545f6b4bee0f4466ef71c7/.travis.yml#L31)? (NB: #886)

PR #873, opened 2023-10-28
make sample watch

vorburger:
- Add make sample
- Add make sample-watch


claat/Makefile
```diff
@@ -61,3 +61,12 @@ $(OUTDIR)/claat-%: $(SRCS) VERSION
 
 %/tmpldata.go: %/gen-tmpldata.go %/template.html %/template.md %/template-offline.html
 	cd $* && go generate
+
+sample: $(OUTDIR)/claat
+	$(OUTDIR)/claat export -o ../sample/ ../sample/codelab.md
+
+sample-watch: sample
+	# This is no good, because serve will keep running in the background even after Ctrl-C:
```
vorburger, line 69: I could probably write a Bash script which remembers the PID and traps the Ctrl-C and then kills it - sort of thing. Is there any better and more direct way?

vorburger: done

Issue #882, opened 2023-10-29
site: npm install: SyntaxError: Missing parentheses in call to 'print'. Did you mean print(...)?


vorburger:
```sh
$ cd site
$ npm install

npm WARN deprecated source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
npm WARN deprecated babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read https://babeljs.io/env to update!
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated trim@0.0.1: Use String.prototype.trim() instead
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5
npm WARN deprecated chokidar@2.1.8: Chokidar 2 does not receive security updates since 2019. Upgrade to chokidar 3 with 15x fewer dependencies
npm WARN deprecated jscs-preset-wikimedia@1.0.1: No longer maintained. We recomment migrating to ESLint with eslint-config-wikimedia.
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated uuid@2.0.3: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
[...]

PR #883, opened 2023-10-29
Test 'npm install' for site/ in GitHub Action

vorburger:
This is intended to (illustrate and later help detect regressions like) #882.

Includes renaming go.yml Github Action to build.yaml.


vorburger:
Once  #865 is merged, this should turn red with failing checks, unless #882 is just a local issue affecting only me.

Please hold merging this until #882 is actually fixed...

.github/workflows/build.yaml
```diff
@@ -30,3 +30,14 @@ jobs:
...
+
+  site:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: NPM Install
+        working-directory: site
+        run: npm install
```
vorburger, line 43: Actually, coming to think of it, given #884 problem, this shouldn't just `npm install` but also test ... what, `./node_modules/.bin/gulp dist` - probably?

